PR #3, opened 2021-07-15
Add Http Support

shekohex:
This will update the relayer architecture to handle both HTTP Requests and WebSockets.

This PR however is not for adding the new Endpoints per Http Relayer Spec, it would only contain the part to support serving HTTP Requests, following PRs would be for adding small functionality and new endpoints.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -362,22 +362,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "async-tungstenite"
-version = "0.13.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


drewstone (approved):
Looks good!

drewstone:
We should consider a docker setup that exposes the right port for running both Http / WS ports. Do we have this already?

shekohex:
> We should consider a docker setup that exposes the right port for running both Http / WS ports. Do we have this already?

Yes, both of them actually are running on the same port, as configured in the config file.
in fact, there is no difference in the behavior, you could check the test file, the only change is the endpoint of the websockets.

PR #4, opened 2021-07-15
Add `api/v1/info` endpoint for relayer info.

shekohex:
This is the following PR after #3 landed in `main`.
this PR adds the following HTTP endpoint: GET `api/v1/info`
this API Endpoint will respond back with relayer configuration.

**Example**:
```json

{

    "substrate": {
        "webb": { },
        "edgeware": null,
        "hedgeware": null,
        "beresheet": null
    },
    "evm": {
        "webb": null,
        "edgeware": null,
        "ganache": {
            "withdrewFee": "0x5",
            "withdrewGaslimit": "0x350000",
            "account": "0x7b1afe2745533d852d6fd5a677f14c074210d896"
        },
        "hedgeware": null,
        "beresheet": {
            "withdrewFee": "0x5",
            "withdrewGaslimit": "0x350000",
            "account": "0x48d41e139d3133f1879ce5080b9c2cb4878332c2"
        },
        "harmony": {
            "withdrewFee": "0x5",
            "withdrewGaslimit": "0x350000",
            "account": "0x48d41e139d3133f1879ce5080b9c2cb4878332c2"
        }
    }

}

```

**Notes**:

* if the network is `null` that obviously means this network is not supported or not configured correctly for this relayer.
* The `account` is derived from the account's private key, this would be used during zkp as the value of the relayer account.


nepoche:
Are there any updates that need to be made to the relayer's config.toml? Using postman, I attempted a GET request for `http://localhost:9955/api/v1/info`, but errored out.

shekohex:
Can you please share the error? also, can you try to open the url in the browser too?

nepoche:
Here are some screenshots of the setup and error:
Screen Shot 2021-07-15 at 4 45 23 PM
Screen Shot 2021-07-15 at 4 45 58 PM
Screen Shot 2021-07-15 at 4 46 32 PM

src/handler.rs
```diff
@@ -79,6 +79,39 @@ pub async fn handle_ip_info(
...
+                .$f
+                .as_mut()
+                .zip(ctx.evm_wallet::<$network>().await.ok())
+            {
+                c.account = Some(w.address());
+            }
+        };
+    }
+
+    update_account_for!(webb, evm::Webb);
```
drewstone, line 106: cool macro! can you document it a bit more?

drewstone (changes requested):
Looks great!

src/handler.rs
shekohex: Done.

tests/main.ts
```diff
@@ -348,6 +348,9 @@ async function main() {
   console.log('Starting the Relayer ..');
   const relayer = await startWebbRelayer();
   await sleep(500); // just to wait for the relayer start-up
+  // get all relayer information
+  const relayerInfoRes = await fetch('http://localhost:9955/api/v1/info');
```
shekohex, line 352: @nepoche I added a way to query relayer information during the tests here, so we at least make sure it works

tests/main.ts
```diff
@@ -399,6 +402,8 @@ async function main() {
     contractAddress
   );
   console.log('Proof Generated!');
+  const relayerChainInfo = relayerInfo.evm[testedChain.name];
+  console.log({ relayerChainInfo });
```
shekohex, line 406: and reading the `withdrewFee` is easy from here.

src/chains.rs
```diff
@@ -118,7 +118,7 @@ pub mod evm {
     }
 
     define_chain! {
-        Harmoney => {
+        Harmony => {
```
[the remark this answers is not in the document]
nepoche: From what I can tell, the shards operate as separate EVMs. So if we are deploying on multiple shards, then we should define separate chains.

PR #16, opened 2021-08-24
Implement geolocation API and support for relayer behind reverse proxy

nepoche:


shekohex:
That's really great, very simple and easy to use.

However, I noticed that you are using the `ipgeolocate` crate, which depends on other _centralized_ IP services, and each service has its own rate limit (on the IP of the caller) so for example, you picked the https://ipapi.co `Service::IpApi` which has a `1,000 queries/day` limit. 
which means that if the dApp got visited 1,000 times (from different users around the globe) that relayer would get rate limited quite easily, and no Ip query can be done from that relayer anymore.

On the other hand, if we let the client (the Webb dApp) call this service instead the caller here would be the user, not the Relayer which in this case have a little to no chance to get rate limited since 1,000 visits from the same user on the same device per day is something really really not possible.

PR #27, opened 2021-09-22
Bridge & Anchor2 Watcher (Phase 2)

shekohex:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [x] Code style update (formatting, local variables)
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

## Other information

Spec: https://www.notion.so/hicommonwealth/Spec-Bridge-Watcher-Phase-2-43cd4a382d2a417eb40953f69685279e


src/events_watcher/bridge_watcher.rs
```diff
@@ -141,11 +174,150 @@ impl EventWatcher for BridgeContractWatcher {
...
+    let block_height_hex = pad32(&format!("{:x}", block_height));
+    let merkle_root_hex = hex::encode(&merkle_root);
+    format!("{}{}{}", chain_id_hex, block_height_hex, merkle_root_hex,)
+}
+
+fn create_resource_id(
+    anchor2_address: types::Address,
+    chain_id: types::U256,
+) -> anyhow::Result<[u8; 32]> {
+    let chain_id_hex = pad(&format!("{:x}", chain_id), 2);
```
drewstone, line 258: We should allow for 4 bytes of data for chain IDs

src/service.rs
```diff
@@ -41,6 +43,10 @@ pub fn ignite(ctx: &RelayerContext, store: Arc<Store>) -> anyhow::Result<()> {
                     )?;
                 }
                 Contract::Bridge(config) => {
+                    let provider = ctx.evm_provider(chain_name).await?;
+                    let wallet = ctx.evm_wallet(chain_name).await?;
+                    let client = SignerMiddleware::new(provider, wallet);
+                    let client = Arc::new(client);
```
drewstone, line 49: Can we combine these? `Arc::new(SignerMiddleware::new(provider, wallet))`

src/events_watcher/bridge_watcher.rs
```diff
@@ -141,11 +174,150 @@ impl EventWatcher for BridgeContractWatcher {
...
+        let status = ProposalStatus::from(status);
+        tracing::debug!("ProposalStatus({:?})", status);
+        tracing::debug!(
+            "Voting with Data = 0x{} & hash = {:?} with resource_id = {:?}",
+            update_data,
+            data_hash,
+            resource_id
+        );
+        // TODO(@shekohex): we should really check if this proposal is active, before sending Tx.
+        // TODO(@shekohex): we need a way to determine whether we should vote
```
drewstone, line 228: Can you create a task for this and link me? These are good small tasks we should triage.

drewstone (changes requested):
looks great, some small things to maybe edit unless i'm mistaken. Also, if we add TODOs, can we create tasks for them as well so we can triage in Notion?

src/events_watcher/bridge_watcher.rs
```diff
@@ -141,11 +174,150 @@ impl EventWatcher for BridgeContractWatcher {
...
+        wrapper: &Self::Contract,
         cmd: BridgeCommand,
     ) -> anyhow::Result<()> {
-        // TODO(@shekohex): Handle the cmd here.
-        tracing::debug!("Got cmd {:?}", cmd);
+        use BridgeCommand::*;
+        tracing::trace!("Got cmd {:?}", cmd);
+        match cmd {
+            CreateProposal(data) => {
+                self.create_proposal(store, &wrapper.contract, data).await?;
```
drewstone, line 185: Is there a way we can have this return the raw transaction? otherwise I would name it `create_and_send_proposal`.

drewstone: If we return raw unsigned transactions and then sign them afterwards, then we can start modularizing for the eventual TX queue.

src/events_watcher/bridge_watcher.rs
shekohex: I guess we need to do the same on the JS side too.
here: https://github.com/webb-tools/darkwebb-solidity/blob/main/test/helpers/bridge.js#L85

shekohex: But can't change it right now, as the tests check against javascript implementation.

shekohex: We should create a spec for the `resoruceId` on notion since I get this error now when I try to create a proposal: `execution reverted: no handler for resourceID`

src/events_watcher/bridge_watcher.rs
shekohex: Yes, that's the plan, we can do that once we get it to work! (a vote tx).

src/events_watcher/bridge_watcher.rs
shekohex: 1. https://www.notion.so/hicommonwealth/Relayer-should-Check-for-Proposal-Status-before-taking-any-action-20c890a280a54174bb54de4c1c311f46
2. https://www.notion.so/hicommonwealth/When-and-How-the-relayer-should-create-a-proposal-64d7074f227b4b0eafec66ba0ebd7f14

src/events_watcher/bridge_watcher.rs
drewstone: It already is correct on JS side.

shekohex: Cool!
Fixed.

PR #44, opened 2021-11-16
Update configuration deserialization to be fault-tolerant of missing env vars

nepoche:
## PR Checklist

Added unused config with undefined environment variable in tests directory

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Users define the networks they wish to support with the appropriate WEBB_EVM_network_ENABLED and network_PRIVATE_KEY environment variables. However, errors still appear for networks they do not intend to support, because the environment variables for private keys for these unsupported networks do not exist.

## What is the new behavior?
Incrementally deserialize each config file, and include the config to the relayer runtime if appropriately parsed. If deserialization fails for whatever reason, skip that config and continue.

- [ ] Yes
- [x] No

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,6 @@ thiserror = "^1.0"
 glob = "^0.3"
 headers = "0.3.5"
 dotenv = "0.15.0"
-regex = "1.5.4"
 serde_path_to_error = "0.1.5"
 
 [dev-dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,7 @@ thiserror = "^1.0"
 glob = "^0.3"
 headers = "0.3.5"
 dotenv = "0.15.0"
+regex = "1.5.4"
 
 [dev-dependencies]
 tempfile = "^3.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "webb-relayer"
-version = "0.1.1"
+version = "0.2.0"
 authors = ["Shady Khalifa <shekohex@gmail.com>"]
 edition = "2018"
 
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -38,6 +38,7 @@ thiserror = "^1.0"
 glob = "^0.3"
 headers = "0.3.5"
 dotenv = "0.15.0"
+regex = "1.5.4"
```
shekohex, line 41: Do we need `regex` now?

src/config.rs
```diff
@@ -252,9 +252,25 @@ pub fn load<P: AsRef<Path>>(path: P) -> anyhow::Result<WebbRelayerConfig> {
     // then get an iterator over all matching files
     let config_files = glob::glob(&pattern)?.flatten();
     for config_file in config_files {
+        let base = config_file.display().to_string();
         let file = config::File::from(config_file).format(FileFormat::Toml);
-        // merge the file into the config
-        cfg.merge(file)?;
+        let mut incremental_config = config::Config::new();
+        incremental_config.merge(file.clone())?;
+        match incremental_config.try_into::<WebbRelayerConfig>() {
```
shekohex, line 259: Could you use `serde_path_to_error::deserialize::<WebbRelayerConfig>(incremental_config)` instead? so we can have a path to the error too here.

src/config.rs
```diff
@@ -252,9 +252,25 @@ pub fn load<P: AsRef<Path>>(path: P) -> anyhow::Result<WebbRelayerConfig> {
...
+                cfg.merge(file)?
+            }
+            Err(e) => {
+                // print the issue that occurred while deserializing, then skip that config
+                tracing::debug!(
+                    "Error {} while attempting to deserialize {}",
+                    e,
+                    base
+                );
+                &cfg
```
shekohex, line 271: why do you return the `cfg` here?

shekohex (changes requested):
Looks good, just left a few changes.

Cargo.toml
nepoche: Removed it

src/config.rs
nepoche: I'm not sure why it required the types to be specified but the code did not have an issue with the later match against config. I updated it with types explicitly stated.

src/config.rs
nepoche: removed

PR #59, opened 2021-12-28
Substrate mixer relay

drewstone:
## PR Checklist

- [ ] Tests for the changes have been added (for bug fixes/features)

## PR Type
This adds the transaction relay functionality for Webb Protocol Substrate-based mixers.

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #57 

- [ ] Yes
- [x] No

### DIFF
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -167,7 +167,7 @@ pub struct MixerRelayTransaction {
     /// The tree id of the mixer's underlying tree
     pub id: u32,
     /// The zero-knowledge proof bytes
-    pub proof: Vec<u8>,
+    pub proof: Bytes,
     /// The target merkle root for the proof
     pub root: Element,
     /// The nullifier_hash for the proof
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -25,12 +25,10 @@ use webb::evm::ethers::{
     signers::{LocalWallet, Signer},
     types::Bytes,
 };
-use webb::substrate::protocol_substrate_runtime::api::{
-    balances::events::BalanceSet, balances::storage::Account,
-    runtime_types::darkwebb_standalone_runtime::Element, AccountData,
-};
-use webb::substrate::subxt;
 
+use webb::substrate::protocol_substrate_runtime::api::runtime_types::darkwebb_standalone_runtime::Element;
+use webb::substrate::subxt::{self, PairSigner, TransactionStatus::{Finalized, InBlock}};
+use webb::substrate::protocol_substrate_runtime::api::{DefaultConfig, RuntimeApi};
 use crate::context::RelayerContext;
 use crate::store::LeafCacheStore;
 
@@ -167,19 +165,19 @@ pub struct MixerRelayTransaction {
     /// The tree id of the mixer's underlying tree
     pub id: u32,
     /// The zero-knowledge proof bytes
-    pub proof: Bytes,
+    pub proof: Vec<u8>,
     /// The target merkle root for the proof
-    pub root: H256,
+    pub root: [u8; 32],
     /// The nullifier_hash for the proof
-    pub nullifier_hash: H256,
+    pub nullifier_hash: [u8; 32],
     /// The receipient of the transaction
     pub recipient: subxt::sp_core::crypto::AccountId32,
     /// The relayer of the transaction
     pub relayer: subxt::sp_core::crypto::AccountId32,
     /// The relayer's fee for the transaction
-    pub fee: U256,
+    pub fee: u128,
     /// The refund for the transaction in native tokens
-    pub refund: U256,
+    pub refund: u128,
 }
 
 #[derive(Debug, Clone, Deserialize)]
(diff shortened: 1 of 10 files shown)


src/handler.rs
```diff
@@ -148,7 +155,32 @@ pub enum Command {
...
+
+#[derive(Debug, Clone, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct MixerRelayTransaction {
+    /// one of the supported chains of this relayer
+    pub chain: String,
+    /// The tree id of the mixer's underlying tree
+    pub id: u32,
+    /// The zero-knowledge proof bytes
+    pub proof: Vec<u8>,
```
shekohex, line 170: Consider using `Bytes` instead of just `Vec<u8>` this would allow the frontend to send it as Hex or Uint8List (a la Byte Buffer).

```suggestion
    pub proof: Bytes,
```

src/handler.rs
```diff
@@ -148,7 +155,32 @@ pub enum Command {
...
+    /// The target merkle root for the proof
+    pub root: Element,
```
shekohex, line 172: The type `Element` will not work here, as it does not implement serde traits.

Prefer the use of `H256` it will give you the exact 32 bytes.
```suggestion
    pub root: H256,
```

src/handler.rs
```diff
@@ -148,7 +155,32 @@ pub enum Command {
...
+    /// The nullifier_hash for the proof
+    pub nullifier_hash: Element,
```
shekohex, line 174: Same as `root`, use `H256`
```suggestion
    pub nullifier_hash: H256,
```

src/handler.rs
```diff
@@ -148,7 +155,32 @@ pub enum Command {
...
+    /// The receipient of the transaction
+    pub recipient: AccountData,
```
shekohex, line 176: This one is a bit tricky here, the way we send it to the verifier on the withdraw is that we use custom encoders to make the `AccountId` into a 32 bytes `Element`. so again you should make it `H256` so the client has to format it, or better use `subxt::sp_core::crypto::AccountId32`
```suggestion
    pub recipient: subxt::sp_core::crypto::AccountId32,
```

src/handler.rs
```diff
@@ -148,7 +155,32 @@ pub enum Command {
...
+    /// The relayer of the transaction
+    pub relayer: AccountData,
```
shekohex, line 178: Same as `recipient` above:
```suggestion
    pub relayer: subxt::sp_core::crypto::AccountId32,
```

src/handler.rs
```diff
@@ -148,7 +155,32 @@ pub enum Command {
...
+    /// The relayer's fee for the transaction
+    pub fee: u128,
```
shekohex, line 180: prefer `U256` over the normal `u128` as it can take bigger numbers plus it can be sent as hex, making life easier when working from the front-end.
```suggestion
    pub fee: U256,
```

src/handler.rs
```diff
@@ -148,7 +155,32 @@ pub enum Command {
...
+    /// The refund for the transaction in native tokens
+    pub refund: u128,
```
shekohex, line 182:
```suggestion
    pub refund: U256,
```

src/handler.rs
```diff
@@ -633,6 +662,85 @@ fn calculate_fee(fee_percent: f64, principle: U256) -> U256 {
...
+                yield Network(NetworkStatus::UnsupportedChain);
+                return;
+            }
+        };
+
+        let api = subxt::ClientBuilder::new()
+            .set_url(chain.http_endpoint)
+            .build()
+            .await?
+            .to_runtime_api::<RuntimeApi<DefaultConfig>>();
```
shekohex, line 698: Try this one is better, as in the future we maybe cache all these clients instead of open a new connection everytime: 
```suggestion
        let client = ctx.substrate_provider::<DefaultConfig>(&requested_chain).await?;
        let api = client.to_runtime_api::<RuntimeApi<DefaultConfig>>();
```

tests/proofUtils.ts
```diff
@@ -235,3 +240,63 @@ export async function generateTornadoSnarkProof(
...
+  const withdrawTx = await nativeTornadoInstance.withdraw(
+    proof, 
+    ...args,
+    {
+      gasLimit: 6000000
+    }
+  );
+  const receipt = await withdrawTx.wait();
+  return receipt;
+}
```
shekohex, line 302: Could you explain what is this function has to do with the substrate? as in it uses ethers here!

tests/integration_tests/substrate/mixerTests.ts
```diff
@@ -0,0 +1,488 @@
...
+    );
+
+    // save the relayer configured parameters
+    calculatedFee = calculateFee(
+      contractConfig?.withdrawFeePercentage ?? 0.0,
+      contractDenomination
+    );
+  });
+
+  describe('Sunny day Tornado Relayed transaction', function () {
```
shekohex, line 62:
```suggestion
  describe('Sunny day Mixer Relayed transaction', function () {
```

shekohex (changes requested):
I left a few comments here and there, also pushed some changes too, that would help to progress this PR.

src/handler.rs
drewstone: But the balance on chain is a u128?

tests/proofUtils.ts
drewstone: Yep, this is unfinished atm.

src/handler.rs
shekohex: I guess you have to think about how you do send it to the withdraw, if it is u128 then yes, it is all good, if it is a 32bytes array, then we need to change it to that.

shekohex (approved):

Issue #73, opened 2022-01-10
[BUG] Subsequent relayer withdraw transactions hit errors


nepoche:
### Overview

While updating integration tests for the relayer to withdraw from subsequent deposits, off of branch `nathan/relayer-subsequent-deposit-issues` the following issue is observed.

### Observed Behavior

` 2022-01-10T13:58:50.546761Z ERROR webb_relayer::tx_queue: Error while sending tx 0x5a5b33473ea4898d5d7ac94414895e19bb5c66b4b316fee615169e8d9c6674ef, (code: -32000, message: VM Exception while processing transaction: revert New leaf index must be greater, data: Some(Object({"name": String("c"), "stack": String("c: VM Exception while processing transaction: revert New leaf index must be greater\n    at Function.c.fromResults (/Users/nathanbarnavon/workspace/relayer/tests/node_modules/ganache-cli/build/ganache-core.node.cli.js:4:192416)\n    at e.exports (/Users/nathanbarnavon/workspace/relayer/tests/node_modules/ganache-cli/build/ganache-core.node.cli.js:55:2089395)")})))`

### Expected Behavior

The relayed withdraw should work successfully

nepoche:
Full logs available here:
[RelayerSubsequentDeposits.txt](https://github.com/webb-tools/relayer/files/7839802/RelayerSubsequentDeposits.txt)

shekohex:
From Reading the Logs you provided, it sounds like there is something wrong with the event watcher, it always get the old events again and again for some reason, which is not good, I'm not sure is the something wrong with ganache itself or we have something in our events watcher regards where we start the blocks.

It is like the following:
```python
while (True):
	step = 50
	start_block = get_last_saved_block()
	last_block = get_last_block_from_chain()
	dest_block = min(start_block + step, last_block)
	events = fetch_events_from_chain(start_block, dest_block)
	process_events(events)
	save_last_block(dest_block)
```

What Happen is, when we start again, it sounds like we got all the events already from the `dest_block` but in the next loop we get the events again from that block, since the `dest_block` now is the `start_block`.

I wonder if we changed the code to be like `save_last_block(dest_block + 1)` would solve this issue, worth trying it @nepoche .

PR #76, opened 2022-01-11
Fix off by one issue in event processing

nepoche:
## PR Checklist

- [ ] Tests for the changes have been added (for bug fixes/features)

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #73 

Events are acted upon twice in the handle_event due to an off-by-one error when querying for a block range..

## What is the new behavior?

Events are only acted upon once.

- [ ] Yes
- [x] No

### DIFF
--- a/tests/integration_tests/ganache/anchorTests.ts
+++ b/tests/integration_tests/ganache/anchorTests.ts
@@ -272,9 +272,9 @@ describe('Anchor Tests', function () {
   });
 
   describe.only('Sunny day Anchor withdraw relayed transaction across bridge', function () {
-    this.timeout(150_000);
+    this.timeout(120_000);
     before(async function () {
-      this.timeout(150_000);
+      this.timeout(120_000);
       sourceWallet = new ethers.Wallet(senderPrivateKey, provider1);
       destWallet = new ethers.Wallet(senderPrivateKey, provider2);
       const chainABridge = bridge.getBridgeSide(chainId1);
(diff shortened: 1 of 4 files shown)


nepoche:
While running the tests for this PR, I came across an error which seems strange: 
After the second deposit, there is a ~30 second gap between signaling the bridge to create a proposal, and actually sending the transaction. These are the logs:
[signalling-proposal-gap.txt](https://github.com/webb-tools/relayer/files/7849223/signalling-proposal-gap.txt)

drewstone:
@nepoche can you start using Gists instead of dumping txt files?

tests/integration_tests/ganache/anchorTests.ts
```diff
@@ -301,11 +302,22 @@ describe('Anchor Tests', function () {
...
 
       // deposit
-      const deposit = await srcAnchor.deposit(chainId2);
+      console.log('before the first deposit');
+      await srcAnchor.deposit(chainId2);
 
       // allow time for the bridge proposal and execution
       console.log('waiting for bridge proposal and execution');
-      await sleep(30_000);
+      await sleep(40_000);
```
drewstone, line 310: Truly this type of addition to the tests does not make it more resilient or allow us to build resiliency into the relayer. In fact, I think it accomplishes the opposite. We need testing tools that "wait for bridge proposal and execution" by doing just that, not by waiting 40 seconds.

drewstone (approved):
LGTM for now. Can you please create tasks to improve the testing flow with utilities to properly wait for actions rather than sleeping for hardcoded times?

shekohex:
Could you change these lines here https://github.com/webb-tools/relayer/blob/main/src/events_watcher/bridge_watcher.rs#L401-L411 by removing the `.block(...)` condition, and see if this make it faster? While this change is much more safer as in one block confirmation is not that bad, but this could be the behavior from ganache itself.

 however since ganache does not mine blocks on itself (there is must a transaction or something to force the chain to mine blocks).

I would suggest improving our tests by forcing ganache to mine a block every 1s for example, or just send `eth_mine` call to the chain to make it progress manually.

Issue #125, opened 2022-04-07
[TASK] Improve test for Anchor and Mixer


dutterbutter:
# Overview

- Check balance of the recipient 
- Check balance of the relayer

dutterbutter:
Closed in #133

PR #133, opened 2022-04-12
include isManual for connecting to protocol substrate node that was m…

### LINKED ISSUE #134: Test should run with Protocol Substrate Node started Manually
One can specify to test `subTransactionRelayer` by first manually starting the protocol substrate node and connecting to the nodes via the ws port.

I suggested this as i could not get my nodes to peer each other when started automatically via the test.

dharjeezy:
connect to protocol substrate node that runs manually.
closes #134 #125

### DIFF
--- a/tests/lib/localProtocolSubstrate.ts
+++ b/tests/lib/localProtocolSubstrate.ts
@@ -16,7 +16,6 @@ export class LocalProtocolSubstrate extends SubstrateNodeBase<TypedEvent> {
     opts: LocalNodeOpts
   ): Promise<LocalProtocolSubstrate> {
     opts.ports = await super.makePorts(opts);
-    console.log(`ports are ${opts.ports}`);
     const startArgs: string[] = [];
     if (opts.usageMode.mode === 'docker') {
       LocalProtocolSubstrate.pullDkgImage({
--- a/tests/lib/localProtocolSubstrate.ts
+++ b/tests/lib/localProtocolSubstrate.ts
@@ -33,13 +33,15 @@ export type NodeOptions = {
     | 'auto';
   authority: 'alice' | 'bob' | 'charlie';
   usageMode: UsageMode;
+  enableLogging?: boolean;
+  isManual?: boolean; // for manual connection to the substrate node using 9944
 };
 
 export class LocalProtocolSubstrate {
   #api: ApiPromise | null = null;
   private constructor(
     private readonly opts: NodeOptions,
-    private readonly process: ChildProcess
+    private readonly process?: ChildProcess,
   ) {}
 
   public get name(): string {
@@ -52,6 +54,14 @@ export class LocalProtocolSubstrate {
     if (opts.ports === 'auto') {
       opts.ports = {
         ws: await getPort({ port: portNumbers(9944, 9999) }),
+        http: await getPort({ port: portNumbers(9933, 9999) }),
+        p2p: await getPort({ port: portNumbers(30333, 30399) }),
+      };
+    }
+
+    if(opts.isManual) {
+      opts.ports = {
+        ws: 9944,
         http: await getPort({ port: portNumbers(9933, 9999) }),
         p2p: await getPort({ port: portNumbers(30333, 30399) }),
       };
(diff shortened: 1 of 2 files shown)


tests/lib/localProtocolSubstrate.ts
```diff
@@ -11,54 +11,12 @@ import {
...
-    if (opts.ports === 'auto') {
-      opts.ports = {
-        ws: await getPort({ port: portNumbers(9944, 9999) }),
-        http: await getPort({ port: portNumbers(9933, 9999) }),
-        p2p: await getPort({ port: portNumbers(30333, 30399) }),
-      };
-    }
-
+    opts.ports = await super.makePorts(opts);
+    console.log(`ports are ${opts.ports}`);
```
shekohex: Remove this:
```suggestion
```

shekohex (changes requested):
Looking good!
just a few changes here.

tests/lib/localProtocolSubstrate.ts
dharjeezy: So, I already removed it.

tests/test/subTransactionRelayer.test.ts
```diff
@@ -99,6 +120,13 @@ describe('Substrate Transaction Relayer', function () {
       relayer: withdrawalProof.relayer,
     });
     expect(txHash).to.be.not.null;
+
+    // get the balance after withdrawal is done and see if it increases
+    // @ts-ignore
+    const { nonce: nonceAfter, data: balanceAfter } = await api.query.system.account(charlie.address);
+    let balanceAfterWithdraw = balanceAfter.free.toBigInt();
+    console.log(`balance after withdrawal is ${balanceAfter.free.toBigInt()}`);
+    expect(balanceAfterWithdraw > initialBalance).true;
```
shekohex, line 129: Try this instead, it is better in terms when the case fails it will print both values for you:
```suggestion
    expect(balanceAfterWithdraw).to.be.gt(initialBalance);
```

tests/test/subTransactionRelayer.test.ts
```diff
@@ -39,6 +39,21 @@ describe('Substrate Transaction Relayer', function () {
...
+      http: 9933,
+      p2p: 30333
+    }
+
+    // for manual connection
+    const bobManualPorts = {
+      ws: 9945,
+      http: 9934,
+      p2p: 30334
+    }
```
shekohex, line 55: Are these are being used? and where?

shekohex (changes requested):
Looks good so far! just a few more comments

tests/test/subTransactionRelayer.test.ts
dharjeezy: These are configurations for the nodes i sprung up manually.

tests/test/subTransactionRelayer.test.ts
dharjeezy: Alright. Thanks

tests/test/subTransactionRelayer.test.ts
```diff
@@ -39,6 +39,21 @@ describe('Substrate Transaction Relayer', function () {
             '../../protocol-substrate/target/release/webb-standalone-node'
           ),
         };
```
shekohex, line 41: You see, usage mode here are still not using the manual starting, am I right?

tests/test/subTransactionRelayer.test.ts
shekohex: But where these variables are being used? can't find a usage for any of them

tests/test/subTransactionRelayer.test.ts
dharjeezy: Hello @shekohex i had to change it back to this `expect(balanceAfterWithdraw > initialBalance).true;`
because the test fails, `balanceAfterWithdraw` and ` initialBalance` are both BigInt and not number or date, therefore `expect(balanceAfterWithdraw).to.be.gt(initialBalance);` fails to assert

tests/test/subTransactionRelayer.test.ts
dharjeezy: Alright. I will remove them then

tests/test/subTransactionRelayer.test.ts
dharjeezy: yh... i will remove the the alice and bob manual ports

shekohex (approved):
Looks good!

PR #142, opened 2022-04-13
Proposals Signing Backends

### LINKED ISSUE #120: [SPEC] More Flexible Signing Backend for Proposals
### Overview
As of now, the current Implementation of handling deposit events would be then handled to construct [`AnchorUpdateProposal`](https://docs.rs/webb-proposals/latest/webb_proposals/struct.AnchorUpdateProposal.html), and then it creates a transaction to be sent the configured DKG node connected to this deposit events watcher which then, later on, will be signed by the DKG.

This is not bad as a start, but it is not that flexible for testing or when you want to test something without having the whole DKG System running.

### Generic Signing backend

we should have something similar to the following:
```rust
#[async_trait::async_trait]
trait SigningBackend {
	type Proposal;
	/// A method to be called first to check if this backend can handle this proposal or not.
	async fn can_handle_proposal(&self, proposal: &Self::Proposal) -> anyhow::Result<bool>;
	/// Send the Unsigned Proposal to the backend to start handling the signing process.
	async fn handle_proposal(&self, proposal: Self::Proposal) -> anyhow::Result<()>;
}
```

And that's it, now the `AnchorWatcher` which is responsible for watching deposit events and creating proposals will be generic over any `B: SigningBackend`, and all it will do after constructing the proposal is call `self.signing_backend.handle_proposal(prop).await?`

### Backends

Right now we have ideas for two different backends so far:

1. **DKGSigningBackend**:
```rust
#[async_trait::async_trait]
impl SigningBackend for DKGSigingBackend {
    type Proposal = webb_proposals::AnchorUpdateProposal;
    async fn can_handle_proposal(&self, proposal: &Self::Proposal) -> anyhow::Result<bool> {
        let resource_id = proposal.header().resource_id();
        let typed_chain_id = resource_id.typed_chain_id();
        let storage_api = self.api.storage().dkg_proposals();
        let maybe_whitelisted = storage_api
            .chain_nonces(TypedChainId::Evm(src_chain_id.as_u32()), None)
            .await?;
        if maybe_whitelisted.is_none() {
            // chain is not whitelisted.
            tracing::warn!(
                "chain {} is not whitelisted, skipping proposal",
                src_chain_id
            );
            return Ok(false);
        }

        let maybe_resource_id = storage_api
            .resources(ResourceId(resource_id.into_bytes()), None)
            .await?;
        if maybe_resource_id.is_none() {
            // resource id doesn't exist.
            tracing::warn!(
                "resource id 0x{} doesn't exist, skipping proposal",
                hex::encode(resource_id.into_bytes())
            );
            return Ok(false);
        }
        // all is good!
        return Ok(true);
    }
    async fn handle_proposal(&self, proposal: Self::Proposal) -> anyhow::Result<()> {
        let tx_api = self.api.tx().dkg_proposals();
        tracing::debug!(
            "sending proposal = nonce: {}, r_id: 0x{}, proposal_data: 0x{}",
            leaf_index,
            hex::encode(resource_id.into_bytes()),
            hex::encode(&proposal.to_bytes())
        );
        let xt = tx_api.acknowledge_proposal(
            Nonce(leaf_index),
            TypedChainId::Evm(src_chain_id.as_u32()),
            ResourceId(resource_id.into_bytes()),
            proposal.to_bytes().into(),
        );
        // TODO: here we should have a substrate based tx queue in the background
        // where just send the raw xt bytes and let it handle the work for us.
        // but this here for now.
        let signer = &self.pair;
        let mut progress = xt.sign_and_submit_then_watch(signer).await?;
        while let Some(event) = progress.next().await {
            tracing::debug!("Tx Progress: {:?}", event);
        }
        return Ok(());
    }
}
```

2. **MockedSigningBackend**:
```rust
#[async_trait::async_trait]
impl SigningBackend for ManualSigningBackend {
    type Proposal = webb_proposals::AnchorUpdateProposal;
    async fn can_handle_proposal(&self, proposal: &Self::Proposal) -> anyhow::Result<bool> {
        // we are always ready, sir 🗿
        return Ok(true);
    }
    async fn handle_proposal(&self, proposal: Self::Proposal) -> anyhow::Result<()> {
        // here we first will load the private key
        // this private key is considered the private key of the Signature Bridge
        // contract governor
        let private_key = self.config.signature_bridge_governor.clone();
        // we need to use ECDSA crate to use the private key and proposal bytes
        // and then generate the signature.
        let proposal_bytes = proposal.to_bytes();
        let signature = ecdsa_signature(&private_key, &proposal_bytes);
        // perfect!
        // we need to signal all the signature bridges in our system with this proposal.
        let bridge_keys = self.webb_config.evm.values().flat_map(|c| {
            c.contracts
                .iter()
                .filter_map(move |contract| match contract {
                    Contract::SignatureBridge(v) => {
                        Some(BridgeKey::new(v.common.address, c.chain_id.into()))
                    }
                    _ => None,
                })
        }); // there is no need to collect here since we can iterate over the keys.

        // now we just signal each bridge with the proposal.
        for bridge_key in bridge_keys {
            tracing::debug!(
                %bridge_key,
                proposal = ?proposal,
                "Signaling Signature Bridge to execute proposal",
            );
            store.enqueue_item(
                SledQueueKey::from_bridge_key(bridge_key),
                BridgeCommand::ExecuteProposalWithSignature {
                    data: proposal_bytes.to_vec(),
                    signature: signature.to_vec(),
                },
            )?;
        }
        Ok(())
    }
}
```

> Feel free to suggest more functionality to be added to this, but in general it should be simple to be implemented.

shekohex:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: Closes #120 

## What is the new behavior?

Here is an example of a configured network to use a DKG node as a signing backend:

```toml
[evm.hermes]
http-endpoint = "http://localhost:5001"
ws-endpoint = "ws://localhost:5001"
chain-id = 5001
private-key = "$HERMES_PRIVATE_KEY"
tx-queue = { max-sleep-interval = 1500 }

[[evm.hermes.contracts]]
contract = "Anchor"
address = "0xD24260C102B5D128cbEFA0F655E5be3c2370677C"
deployed-at = 1
size = 1
events-watcher = { enabled = true, polling-interval = 1000, print-progress-interval = 0 }
withdraw-fee-percentage = 0
withdraw-gaslimit = "0x350000"
linked-anchors = [
  { chain = "athena", address = "0xD30C8839c1145609E564b986F667b273Ddcb8496" },
]
# Notice here, that we can define what kind of SigningBackend we should use.
# For DKGNode for example, we have to define what is the name of the node that we defined on our config
# somewhere else.
proposal-signing-backend = { type = "DKGNode", node = "dkg-local" }

[[evm.hermes.contracts]]
contract = "SignatureBridge"
address = "0xDe09E74d4888Bc4e65F589e8c13Bce9F71DdF4c7"
deployed-at = 1
events-watcher = { enabled = true, polling-interval = 1000, print-progress-interval = 0 }
```

Here is another example that uses the Mocked Signing Backend:

```toml
[evm.athena]
http-endpoint = "http://localhost:5002"
ws-endpoint = "ws://localhost:5002"
chain-id = 5002
private-key = "$ATHENA_PRIVATE_KEY"
tx-queue = { max-sleep-interval = 1500 }

[[evm.athena.contracts]]
contract = "Anchor"
address = "0xD30C8839c1145609E564b986F667b273Ddcb8496"
deployed-at = 1
size = 1
events-watcher = { enabled = true, polling-interval = 1000, print-progress-interval = 0 }
withdraw-fee-percentage = 0
withdraw-gaslimit = "0x350000"
linked-anchors = [
  { chain = "hermes", address = "0xD24260C102B5D128cbEFA0F655E5be3c2370677C" },
]
# Notice in this example, however, we are not using any DKG node, instead we are using a mocked
# backend that will sign the proposals using the Governor private key directly, and send the signature
# to the signature bridge, to execute the proposal.
proposal-signing-backend = { type = "Mocked", private-key = "$GOVERNOR_PRIVATE_KEY" }

[[evm.athena.contracts]]
contract = "SignatureBridge"
address = "0x2946259E0334f33A064106302415aD3391BeD384"
deployed-at = 1
events-watcher = { enabled = true, polling-interval = 1000, print-progress-interval = 0 }
```

- [x] Yes
- [ ] No

Config Changes described above.

## Other information

See Spec. #120


src/events_watcher/signing_backend/mod.rs
```diff
@@ -0,0 +1,18 @@
+#[doc(hidden)]
+mod dkg;
+#[doc(hidden)]
+mod mocked;
+
+pub use dkg::*;
+pub use mocked::*;
```
dutterbutter, line 7: It would be good to document these for the implementation docs.

dutterbutter (approved):
LGTM! 

Can we just add a row to the config table in the `README.md` for the `signing-backend` field?

src/config.rs
```diff
@@ -314,14 +321,14 @@ pub struct AnchorContractConfig {
     /// Anchor withdraw configuration.
     #[serde(flatten)]
     pub withdraw_config: AnchorWithdrawConfig,
-    /// The name of the DKG node that this contract will use.
-    ///
-    /// Must be defined in the config.
-    pub dkg_node: String,
+    /// The type of the Siging backend used for signing proposals.
```
drewstone, line 324: Spelling

drewstone (changes requested):
What about calling this the ProposalSigningBackend?

I'm thinking about when we will integrate signing of txes for other chains and whether this name can be confused for those signing backends. Thoughts?

tests/lib/localTestnet.ts
```diff
@@ -178,35 +194,31 @@ export class LocalChain {
...
+    const otherAnchors = otherChainIds.map((chainId) =>
+      bridge.getAnchor(chainId, ethers.utils.parseEther('1'))
     );
 
     const chainInfo: FullChainInfo = {
       enabled: true,
       httpEndpoint: this.endpoint,
       wsEndpoint: this.endpoint.replace('http', 'ws'),
-      chainId: this.opts.chainId,
+      chainId: this.underlayingChainId,
```
dharjeezy, line 221: should be underlyingChainId, yh?

tests/lib/localTestnet.ts
```diff
@@ -87,6 +99,10 @@ export class LocalChain {
   }
 
   public get chainId(): number {
+    return Utility.getChainIdType(this.opts.chainId);
+  }
+
+  public get underlayingChainId(): number {
```
dharjeezy, line 105: should be underlyingChainId

src/config.rs
shekohex: Fixed.

drewstone (approved):
LGTM

PR #144, opened 2022-04-14
negative test for Substrate transaction relayer

### LINKED ISSUE #145: [TASK] Add additional negative tests for Substrate Anchor
### Test Cases

- As a relayer I should reject a relayer address that is not my beneficiary
- As a relayer I should reject an invalid proof
- As a relayer I should reject a fee that is less than requested

dharjeezy:
Negative tests for substrate Anchor

closes #145

### DIFF
--- a/tests/test/subTransactionRelayer.test.ts
+++ b/tests/test/subTransactionRelayer.test.ts
@@ -44,14 +44,14 @@ describe('Substrate Transaction Relayer', function () {
       name: 'substrate-alice',
       authority: 'alice',
       usageMode,
-      ports: 'auto',
+      ports: 'auto'
     });
 
     bobNode = await LocalProtocolSubstrate.start({
       name: 'substrate-bob',
       authority: 'bob',
       usageMode,
-      ports: 'auto',
+      ports: 'auto'
     });
 
     await aliceNode.writeConfig({
@@ -71,8 +71,10 @@ describe('Substrate Transaction Relayer', function () {
   });
 
   it('Simple Mixer Transaction', async () => {
-    const withdrawalProof = await initWithdrawal(aliceNode, webbRelayer, '//Charlie');
     const api = await aliceNode.api();
+    const account = createAccount('//Charlie');
+    const note = await makeDeposit(api, aliceNode, account);
+    const withdrawalProof = await initWithdrawal(api, webbRelayer, account, note);
     // get the initial balance
     // @ts-ignore
     let { nonce, data: balance } = await api.query.system.account(withdrawalProof.recipient);
@@ -101,7 +103,10 @@ describe('Substrate Transaction Relayer', function () {
   });
 
   it('Should fail to withdraw if address is invalid', async () => {
-    const withdrawalProof = await initWithdrawal(aliceNode, webbRelayer, '//Dave');
+    const api = await aliceNode.api();
+    const account = createAccount('//Dave');
+    const note = await makeDeposit(api, aliceNode, account);
+    const withdrawalProof = await initWithdrawal(api, webbRelayer, account, note);
 
     const invalidAddress = "5DAAnrj7VHTznn2AWBemMuyBwZWs6FNFjdyVXUeYum3PTXFy";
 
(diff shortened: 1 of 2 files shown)


tests/test/subTransactionRelayer.test.ts
```diff
@@ -235,3 +301,26 @@ async function createMixerWithdrawProof(
...
+  const account = keyring.addFromUri(accountId);
+  // send the deposit transaction.
+  const txSigned = tx.sign(account);
+  await aliceNode.executeTransaction(txSigned);
+  // next we need to prepare the withdrawal transaction.
+  // create correct proof with right address
+  const withdrawalProof = await createMixerWithdrawProof(api, note, {
+    recipient: account.address,
+    relayer: account.address,
+    fee: 200,
```
drewstone, line 318: Can we remove these hardcoded values and provide them as optional parameters? Default should be 0 in my opinion.

tests/test/subTransactionRelayer.test.ts
```diff
@@ -71,26 +71,11 @@ describe('Substrate Transaction Relayer', function () {
   });
 
   it('Simple Mixer Transaction', async () => {
+    const withdrawalProof = await initWithdrawal(aliceNode, webbRelayer, '//Charlie');
```
drewstone, line 74: I think using helper functions here hurts the clarity and readability of the test. This function seems to do 2 things:
1. Generates and executes a deposit transaction.
2. Creates a withdrawal proof from that deposit.

I think given those are 2 different steps, can we consider using 2 separate helper functions here for them? It should be clear to the test reader how the system works in a simpler fashion than just wrapping all shared logic into a function. What do you think?

drewstone (changes requested):

tests/test/subTransactionRelayer.test.ts
dharjeezy: Alright. Will do

tests/test/subTransactionRelayer.test.ts
dharjeezy: Alright. I can sure decouple the helper function

tests/test/subTransactionRelayer.test.ts
```diff
@@ -109,12 +94,93 @@ describe('Substrate Transaction Relayer', function () {
...
+        proof: Array.from(hexToU8a(withdrawalProof.proofBytes)),
+        root: Array.from(hexToU8a(withdrawalProof.root)),
+        nullifierHash: Array.from(hexToU8a(withdrawalProof.nullifierHash)),
+        refund: withdrawalProof.refund,
+        fee: invalidFee,
+        recipient: withdrawalProof.recipient,
+        relayer: withdrawalProof.relayer,
+      });
+    } catch (e) {
+      console.log(`error is ${e}`);
```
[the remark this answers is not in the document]
dharjeezy: Alright. Will do.

tests/test/subTransactionRelayer.test.ts
```diff
@@ -109,12 +94,93 @@ describe('Substrate Transaction Relayer', function () {
...
+
+      // Expect an error to be thrown
+      expect(e).to.not.be.null
+    }
+  });
+
+  it('Should fail to withdraw if proof is invalid', async () => {
+    const withdrawalProof = await initWithdrawal(aliceNode, webbRelayer, '//Eve');
+
+    const invalidProofByte = "0x4243ba7b4f0fc4aac6e6d02f0810a8413df56269d6407786192f8d5da8dc32880acca580f677896ebba7704aec84ba5cfd634af254995cdc985bf2c26b5f43171f365b3b8e03fc833e67e6640e9a7e048e62ce4982d39b23d091cd00a584679f34142063ce11555a21d150c38d144109e21f09d6761c939c8462ab72ae0b420e";
```
shekohex, line 133: I would say instead of hardcode an invalid proof here as a long hex string, I would say get the original proof, convert it into `Uint8Array`, and then change the first byte to something else:
```js
const proofBytes = hexToU8a(originalProofBytes);
proofBytes[1] |= 0x42;
const invalidProofBytes = u8aToHex(proofBytes);
expect(originalProofBytes).to.not.eq(invalidProofBytes);
```

Something like that.

shekohex (changes requested):

tests/test/subTransactionRelayer.test.ts
dharjeezy: Alright. That also works

PR #158, opened 2022-04-20
Substrate anchor relayer support

### LINKED ISSUE #152: [TASK] Substrate anchor relayer support
# Overview
We have support for the Substrate mixer, therefore it shouldn't be terribly different to integrate support for the Substrate anchor.

# Checklist
- [x] Add a handler for the Substrate Anchor, using the Substrate mixer handler as a reference.
- [x] Add a test for it as an integration test.
- [ ] Add inline documentation to the file and wherever else is relevant if there's missing docs.

dharjeezy:
closes #152


drewstone:
Lints and clippy fixes needed. Also I think we could improve the anchor test to get the roots from the chain directly. It would make it clearer how roots are meant to be gathered when making proofs.

PR #171, opened 2022-05-02
more negative tests

dharjeezy:
More negative tests


shekohex (approved):
LGTM!

We will ignore the CI for now, until we use another free bigger CI Machines.

PR #178, opened 2022-05-10
Mocked Signing Backend Generalization

### LINKED ISSUE #175: [TASK] Generalize SignatureBridge target system for mocked signing backends
# Overview
Currently the mocked backend uses a SignatureBridgeMetadata that is tied to EVM systems. We want to ensure this works over arbitrary chains, namely Substrate, CosmWasm, and eventually Ink. We will likely want to leverage the target system struct for this.

## What exists
```rust
pub struct SignatureBridgeMetadata {
    pub chain_id: TypedChainId,
    pub address: Address,
    pub private_key: PrivateKey,
}
```

## What we should investigate
```rust
pub struct SignatureBridgeMetadata {
    pub chain_id: TypedChainId,
    pub address: TargetSystem,
    pub private_key: PrivateKey,
}
```

## In the mocked backend
In the mocked backend, we should consider creating the proposals for each target chain on the fly, using the same raw `AnchorUpdateProposal` for example. It could like like the following but this does not necessarily cover all cases.
```rust
let proposal_bytes match dest_chain_id {
    TypedChainId::Evm(id) => { ... },
    TypedChainId::Cosmos(id) => { ... },
    TypedChainId::Substrate(id) => { ... },
}
```

shekohex:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: Closes #175 

## What is the new behavior?

Now the Mocked Signing Backend is general over any system or any target chain, so it can be used not only in the EVM Chains context but also (and not limited to) Substrate systems.

- [x] Yes
- [ ] No

I would say **Yes** here since from now on when you going to use the Mocked Signing Backend, the whole system must share the same Governor, which means on all systems (`SignatureBridge{Contract,Pallet,Program,...}`) must all use the same governor identity/public key.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -148,9 +148,9 @@ dependencies = [
 
 [[package]]
 name = "autocfg"
-version = "1.0.1"
+version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cdb031dd78e28731d87d56cc8ffef4a8f36ca26c38fe2de700543e627f8a464a"
+checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
 [[package]]
 name = "backoff"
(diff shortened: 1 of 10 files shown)


salman01zp (approved):
LGTM..!! Great Work

PR #182, opened 2022-05-16
VAnchor Governance

### LINKED ISSUE #180: [TASK] Listen for VAnchor events
### Overview

Similar to the normal `Anchor` contract watcher, for governance, we need to implement the same for the `VAnchor` contract.

The original `Anchor` governance logic is implemented [here](https://github.com/webb-tools/relayer/blob/9f4bfd25c22209cbb723f1c77205c898db6292d7/src/events_watcher/evm/anchor_watcher.rs).

### Anchor Governance Flow

In the `Anchor` contract (and similarly will be for `VAnchor`) when we see a `Deposit` event, we do the following:

1. Go through every Linked Anchor to this contract.
2. Construct `AnchorUpdateProposal` from the deposit information.
3. Check if the `ProposalSigningBackend` can sign it and if so we hand over the proposal.
4. Then we mark this deposit as processed.

That's it, this is all we need from this service.

> In the background, the signing backend will sign this proposal, and eventually, this proposal will go to the targeted `SignatureBridge`.

### VAnchor Governance Flow

Very Similar to the regular Anchor governance flow except for the following:

1. Only construct the `AnchorUpdateProposal` if this condition evaluates to `true`: `leaf_index % 2 != 0`

The reason behind this is that `VAnchor` on every `transact` call, emits two events, similar to the `Deposit` event but we call it the `NewCommitment` event, a la two `UTXO` and since we only need to update the target `VAnchor` only when needed, the first `NewCommitment` event sounds redundant in this case.

And to be clear, here is the final flow for the `VAnchor` governance part:

1. Check `leaf_index % 2 != 0` is `true`, if not we skip this event.
2. Go through every Linked Anchor to this contract.
3. Construct `AnchorUpdateProposal` from the deposit information.
4. Check if the `ProposalSigningBackend` can sign it and if so we hand over the proposal.
5. Then we mark this deposit as processed.

shekohex:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: Closes #180 

- [ ] Yes
- [x] No

### DIFF
--- a/tests/test/evm/vanchorGovernance.test.ts
+++ b/tests/test/evm/vanchorGovernance.test.ts
@@ -1,0 +1,251 @@
+/*
+ * Copyright 2022 Webb Technologies Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+import { expect } from 'chai';
+import { Tokens, VBridge } from '@webb-tools/protocol-solidity';
+import { Utxo } from '@webb-tools/utils';
+import { ethers, BigNumber } from 'ethers';
+import temp from 'temp';
+import { LocalChain } from '../../lib/localTestnet.js';
+import { EnabledContracts, WebbRelayer } from '../../lib/webbRelayer.js';
+import getPort, { portNumbers } from 'get-port';
+import { u8aToHex } from '@polkadot/util';
+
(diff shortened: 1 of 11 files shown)


tests/test/evm/vanchorGovernance.test.ts
```diff
@@ -0,0 +1,251 @@
...
+    await token.mintTokens(wallet1.address, ethers.utils.parseEther('1000'));
+    const webbBalance = await token.getBalance(wallet1.address);
+    expect(webbBalance.toBigInt() > ethers.utils.parseEther('1').toBigInt()).to
+      .be.true;
+    const dummyInputs = Array(2)
+      .fill(null)
+      .map(
+        () =>
+          new Utxo({
+            originChainId: BigNumber.from(localChain1.chainId),
```
drewstone, line 194: These should by typed chain IDs

drewstone (changes requested):
LGTM just need to figure out failing CI and small fixes to chain IDs so our examples are consistent.

tests/test/evm/vanchorGovernance.test.ts
shekohex: They are already TypedChainId, if you want the original one you call `underlyingChainId`.

drewstone:
Ok!

PR #200, opened 2022-06-16
update relayer configuration

### LINKED ISSUE #195: [TASK] Separate config requirements for each relayer component/sub-system
### Overview

The relayer system is composed of three main components. Each of these components should be thought of as entirely separate because they could be handled by different entities entirely.

1. Private transaction relaying 
2. Data querying (for zero-knowledge proof generation)
3. Event listening, proposing, and signature relaying (of DKG proposals where the relayer acts like an oracle)

However, currently users are unable to run each component separately or distinctively. For example, the configuration requirements for governance relaying also makes the relayer available for private tx relaying and data querying. We need to modify the configuration options so that users who ONLY want to participate in governance relaying can do so without also relaying private txs / data querying. 

### Use stories

- [ ] As a user, I want to operate a relayer so that I can relay transactions privately. 
- [ ] As a user, I want to operate a relayer so that I can participate in governance relaying.
- [ ] As a user, I want to operate a relayer so that I can participate in zero-knowledge proof generation and serving up data. 

## SPEC
Adding `features`  struct to `WebbRelayerConfig`  which can be used to configure the relayer with different options. By default, all features are enabled.
**Note** :  In order to run relayer for data querying, relayer should also be configured for governance relaying
                   Data querying option will only make leaves API endpoint accessible.

```rust
/// FeaturesConfig is the configuration for running relayer with option.
#[derive(Debug, Clone, Copy, Deserialize, Serialize, Default)]
pub struct FeaturesConfig {
    /// Enable data quering for leaves
    pub data_query: bool,
    /// Enable governance relaying
    pub governance_relay: bool,
    /// Enable private tx relaying
    pub private_tx_relay: bool,
}

pub struct WebbRelayerConfig {
    #[serde(default)]
    pub evm: HashMap<String, EvmChainConfig>,
    #[serde(default)]
    pub substrate: HashMap<String, SubstrateConfig>,
    /// Configuration for running relayer
    ///
    /// by deafult all features are enabled
    /// Features:
    /// 1. Data quering for leaves
    /// 2. Governance relaying
    /// 3. Private transaction relaying
    #[serde(default)]
    pub features: FeaturesConfig,
}
```
### Example
Here is an example of how to configure a relayer for relaying private tx for evm
```rust
port = 9955
.
.
[features]
data-query = false
governance-relay = false
```

### LINKED ISSUE #225: [SPEC] Split the `EventWatcher` trait in two traits `EventWatcher` and `EventHandler`
## Overview

The current [`EventWatcher`](https://github.com/webb-tools/relayer/blob/3eaf5e26d8ac69c7d21d125311d8cf517dda1652/src/events_watcher/mod.rs#L84) does two things:

* [`run`](https://github.com/webb-tools/relayer/blob/3eaf5e26d8ac69c7d21d125311d8cf517dda1652/src/events_watcher/mod.rs#L109): which is the main point of that trait, that runs indefinitely in the background and pulls new events for the configured contract, and this function is _implemented by default_ and the user of this trait does not need to implement this method.
* [`handle_event`](https://github.com/webb-tools/relayer/blob/3eaf5e26d8ac69c7d21d125311d8cf517dda1652/src/events_watcher/mod.rs#L92): Which is the method that needs to be implemented by the Event Watcher to react/handle that event.

This was ideal and works good, However, for only one contract and one event watcher, if we need to watch the events for the same contract but do something else, we need to construct another or a separate event watcher which means we will pull the events twice from the network.

## Goal

- The goal is to watch for an event at most ONCE and be able to handle it MULTIPLE times, instead of watching for the same event MULTIPLE times and handling it MULTIPLE times.
- The goal is to be able to implement multiple handlers for an event watcher and keep it separated.

In a simpler way to describe this, is that we need to have only one `EventWatcher` per contract and many `EventHandler` that are injected in that Event Watcher. Here is how it should look like:

```rust
/// A trait for watching events from a watchable contract.
/// EventWatcher trait exists for deployments that are smart-contract / EVM based
#[async_trait::async_trait]
pub trait EventWatcher {
    const TAG: &'static str;
    type Middleware: providers::Middleware + 'static;
    type Contract: Deref<Target = contract::Contract<Self::Middleware>>
        + WatchableContract;
    type Events: contract::EthLogDecode;
    type Store: HistoryStore + EventHashStore;
	type EventHandler: EventHandler;

    async fn run(
        &self,
        client: Arc<Self::Middleware>,
        store: Arc<Self::Store>,
        contract: Self::Contract,
    ) -> anyhow::Result<()> {
			// Run a loop where it polls the events for each block
			// and then sends each event to the `handle_event` method on each event handler.
			// if it matches the filters.
	}
}

```

```rust
/// A trait for watching events from a watchable contract.
/// EventWatcher trait exists for deployments that are smart-contract / EVM based
#[async_trait::async_trait]
pub trait EventHandler {
	const TAG: &'static str;
    type Middleware: providers::Middleware + 'static;
    type Contract: Deref<Target = contract::Contract<Self::Middleware>>
        + WatchableContract;
    type Events: contract::EthLogDecode;
    type Store: HistoryStore + EventHashStore;

	async fn handle_event(
        &self,
        store: Arc<Self::Store>,
        contract: &Self::Contract,
        (event, log): (Self::Events, contract::LogMeta),
    ) -> anyhow::Result<()>;
}
```

Then the usage would be simply plug in the `EventHandler`s like so:

```rust
pub struct LeavesHandler;
pub struct AnchorUpdateHandler;

impl EventHandler for LeavesHandler {

	async fn handle_event(
        &self,
        store: Arc<Self::Store>,
        contract: &Self::Contract,
        (event, log): (Self::Events, contract::LogMeta),
    ) -> anyhow::Result<()> { ... }
}

impl EventHandler for AnchorUpdateHandler {
		async fn handle_event(
        &self,
        store: Arc<Self::Store>,
        contract: &Self::Contract,
        (event, log): (Self::Events, contract::LogMeta),
    ) -> anyhow::Result<()> { ... }
}

impl EventWatcher for VAnchorWatcher {
	const TAG: &'static str = "VAnchor Watcher";
    type Middleware = HttpProvider;
    type Contract = VAnchorContractWrapper<Self::Middleware>;
    type Events = VAnchorContractEvents;
    type Store = SledStore;
	type EventHandler = (LeavesHandler, AnchorUpdateHandler)
    ...
}
```

salman01zp:
## Task Checklist
Relayer configurations
- [x]  Private transaction relaying.
- [x]  Governance Relaying
- [x]  Data querying

## PR Checklist

**Reference issue to close (if applicable)**

Closes #195


salman01zp:
will skip tests using the DKG backend, until we merge the async branch

src/config.rs
```diff
@@ -744,5 +773,12 @@ fn postloading_process(
             }
         }
     }
+    // check if data quering is enabled then governance_relaying should also be enabled
+    if config.features.data_query && !config.features.governance_relay {
+        tracing::warn!(
+            "!!WARNING!!: In order to enable data quering for relayer,
+            governance relaying must be configured."
+        );
+    }
```
shekohex, line 782: I do think these two options are related, You can enable the data query for helping the people who are using the dApp to be able to generate zKP quickly, However, enabling this does not mean I want to participate in the governance relaying. 

Instead, when I enable the `governance_relay` but I did not enable the leaves watcher for this specific contract, it does not make sense, as we want to watch for the events to be able to work as a governance relayer, the same goes for the case when I enable the `data_query` but I did not enable the `leaves_watcher`.

src/handler.rs
```diff
@@ -257,10 +259,23 @@ pub async fn handle_leaves_cache(
...
+    // check if data query is enabled for relayer
+    if !is_data_query_enabled {
+        tracing::warn!("Data query is not enabled for relayer.");
+        return Ok(warp::reply::with_status(
+            warp::reply::json(&String::from(
+                "Data query is not enabled for relayer.",
+            )),
+            warp::http::StatusCode::FORBIDDEN,
+        ));
+    }
```
shekohex, line 271: This should be a `struct UnsupportedFeature { message: String }` instead of the raw json string.

src/handler.rs
```diff
@@ -337,12 +352,21 @@ pub async fn handle_cmd(
...
+            Command::Evm(evm) => handle_evm(ctx, evm, stream).await,
+            Command::Ping() => {
+                let _ = stream.send(Pong()).await;
+            }
         }
+    } else {
+        tracing::error!("Private transaction relaying is not configured..!");
+        let _ = stream
+            .send(Error(
+                "Private transaction relaying is not configured..!".to_string(),
```
shekohex, line 367:
```suggestion
                "Private transaction relaying is not enabled.".to_string(),
```

src/service.rs
```diff
@@ -79,6 +79,12 @@ pub async fn ignite(
     ctx: &RelayerContext,
     store: Arc<Store>,
 ) -> anyhow::Result<()> {
+    tracing::debug!("Relayer configuration  : {:?}", ctx.config);
+    // check if relayer is configured for governance relaying
+    if !ctx.config.features.governance_relay {
+        tracing::error!("Governance relaying is not configured.");
+        return Ok(());
+    }
```
shekohex, line 87: Services are not only for governance relayer, they are also for leaves watchers, which are different than the governance relayer services.

shekohex (changes requested):

drewstone (changes requested):

PR #227, opened 2022-07-20
Dkg Demo

nepoche:
## Summary of changes
- Use the present structures in the relayer to demonstrate communication between evm chains and dkg.


shekohex:
We made a better demo after that one, so I'm closing this PR.

PR #228, opened 2022-07-21
Events Watcher API 2.0

### LINKED ISSUE #225: [SPEC] Split the `EventWatcher` trait in two traits `EventWatcher` and `EventHandler`

shekohex:
## Summary of changes
- [x] New Events Watcher API For EVM
- [ ] New Events Watcher API For Substrate

- Closes #225 

-----
### Code Checklist 

- [x] Tested
- [x] Documented

### DIFF
--- a/src/events_watcher/evm/signature_bridge_watcher.rs
+++ b/src/events_watcher/evm/signature_bridge_watcher.rs
@@ -26,7 +26,7 @@ use webb::evm::ethers::types;
 use webb::evm::ethers::utils;
 
 use crate::config;
-use crate::events_watcher::{BridgeWatcher, EventWatcher};
+use crate::events_watcher::{BridgeWatcher, EventHandler, EventWatcher};
 use crate::store::sled::{SledQueueKey, SledStore};
 use crate::store::{BridgeCommand, QueueStore};
 
@@ -88,6 +88,9 @@ impl<M: Middleware> super::WatchableContract
 #[derive(Copy, Clone, Debug, Default)]
 pub struct SignatureBridgeContractWatcher;
 
+#[derive(Copy, Clone, Debug, Default)]
+pub struct SignatureBridgeGovernanceOwnershipTransferredHandler;
+
 #[async_trait::async_trait]
 impl EventWatcher for SignatureBridgeContractWatcher {
     const TAG: &'static str = "Signature Bridge Watcher";
@@ -99,7 +102,17 @@ impl EventWatcher for SignatureBridgeContractWatcher {
     type Events = SignatureBridgeContractEvents;
 
     type Store = SledStore;
+}
 
+#[async_trait::async_trait]
+impl EventHandler for SignatureBridgeGovernanceOwnershipTransferredHandler {
+    type Middleware = HttpProvider;
+
+    type Contract = SignatureBridgeContractWrapper<Self::Middleware>;
+
+    type Events = SignatureBridgeContractEvents;
+
+    type Store = SledStore;
     #[tracing::instrument(
         skip_all,
         fields(event_type = ?e.0),
--- a/src/events_watcher/mod.rs
+++ b/src/events_watcher/mod.rs
@@ -78,6 +78,16 @@ pub trait WatchableContract: Send + Sync {
     fn print_progress_interval(&self) -> Duration;
 }
 
+pub type EventHandlerFor<W> = Box<
+    dyn EventHandler<
+            Middleware = <W as EventWatcher>::Middleware,
+            Contract = <W as EventWatcher>::Contract,
+            Events = <W as EventWatcher>::Events,
+            Store = <W as EventWatcher>::Store,
+        > + Send
+        + Sync,
+>;
+
 /// A trait for watching events from a watchable contract.
 /// EventWatcher trait exists for deployments that are smart-contract / EVM based
 #[async_trait::async_trait]
@@ -88,14 +98,6 @@ pub trait EventWatcher {
         + WatchableContract;
     type Events: contract::EthLogDecode;
     type Store: HistoryStore + EventHashStore;
-
-    async fn handle_event(
-        &self,
-        store: Arc<Self::Store>,
-        contract: &Self::Contract,
-        (event, log): (Self::Events, contract::LogMeta),
-    ) -> anyhow::Result<()>;
-
     /// Returns a task that should be running in the background
     /// that will watch events
     #[tracing::instrument(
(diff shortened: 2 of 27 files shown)


src/events_watcher/mod.rs
```diff
@@ -154,13 +154,20 @@ pub trait EventWatcher {
...
-                                store.clone(),
-                                &contract,
-                                (event, log.clone()),
-                            )
-                            .await;
+                        // try to handle the event with all the handlers, simultaneously.
+                        //
+                        // if any of the handlers fail, it will cancel the rest of the handlers.
+                        // and we will retry the event again.
+                        let result = futures::future::try_join_all(
```
drewstone, line 161: Let's instead process all futures and simply retry the ones that fail some constant # of times. We should not cancel the rest of the handlers when one fails.

src/probe.rs
```diff
@@ -37,4 +37,7 @@ pub enum Kind {
     /// Relayer Leaves Store state on a specific chain/node.
     #[display(fmt = "leaves_store")]
     LeavesStore,
+    /// When the Relayer sees a new Merkle Tree insertation event on a specific chain/node.
+    #[display(fmt = "mt_insert")]
+    MerkleTreeInsertation,
```
drewstone, line 42:
```suggestion
    MerkleTreeInsertion,
```

drewstone: Please fix other references to this enum

drewstone (changes requested):
Needs fixes, I'm not confident seeing how you're handling the futures of handlers that we are matching our expectations. Can you provide more info on why you think we should `try_join_all`? Why would we want to cancel futures if one fails?

src/probe.rs
shekohex: Done.

src/events_watcher/mod.rs
shekohex: I addressed your comments here too.

drewstone (approved):

src/events_watcher/evm/signature_bridge_watcher.rs
```diff
@@ -88,6 +88,9 @@ impl<M: Middleware> super::WatchableContract
 #[derive(Copy, Clone, Debug, Default)]
 pub struct SignatureBridgeContractWatcher;
 
+#[derive(Copy, Clone, Debug, Default)]
+pub struct SignatureBridgeGovernanceOwnershipTransferredHandler;
```
salman01zp, line 92: can we rename this struct to something small?

shekohex: It is more like each handler name now at least should describes what events that they are handing. But nonetheless, I'm open for a better name.

salman01zp: nothing comes to mind at present. Anyways we can update later.

salman01zp (approved):
LGTM..! Great work @shekohex

PR #230, opened 2022-07-26
Add the `mnemonic` type for the cosmwasm relayer

duguorong009:
## Summary of changes
- Introduce the new type `mnemonic` for **cosmwasm relayer implementation**

- Closes 

-----
### Code Checklist 

- [ ] Tested
- [ ] Documented


salman01zp (commented):
LGTM..! Just one question

I think you were also going to create `CommonContractConfig` for cosmwasm [here](https://github.com/webb-tools/relayer/blob/main/src/config.rs#L343-L352) due to incompatible address type between evm and cosmwasm

duguorong009:
> LGTM..! Just one question
> 
> I think you were also going to create `CommonContractConfig` for cosmwasm [here](https://github.com/webb-tools/relayer/blob/main/src/config.rs#L343-L352) due to incompatible address type between evm and cosmwasm

Yes, you are right @salman01zp 
I got the similar tips from @shekohex

src/config.rs
```diff
@@ -37,9 +38,9 @@ use ethereum_types::{Address, U256};
 use serde::{Deserialize, Serialize};
 use webb::substrate::subxt::sp_core::sr25519::Public;
 
-use crate::types::private_key::PrivateKey;
 use crate::types::rpc_url::RpcUrl;
 use crate::types::suri::Suri;
+use crate::types::{mnemonic::Mnemonic, private_key::PrivateKey};
```
shekohex, line 43:
```suggestion
use crate::types::{rpc_url::RpcUrl, suri::Suri, mnemonic::Mnemonic, private_key::PrivateKey};
```

src/types/mnemonic.rs
```diff
@@ -0,0 +1,79 @@
+use serde::Deserialize;
+
+/// Mnemonic represents a mnemonic.
+#[derive(Clone)]
+pub struct Mnemonic(Vec<String>);
```
shekohex, line 5: Could we use this https://docs.rs/tiny-bip39/latest/bip39/struct.Mnemonic.html#method.from_phrase or https://docs.rs/tiny-bip39/latest/bip39/struct.Mnemonic.html#method.validate for Mnemonic validation?

src/types/mnemonic.rs
```diff
@@ -0,0 +1,79 @@
...
+                    })?;
+                    str_value = val;
+                } else {
+                    str_value = value.to_string();
+                }
+                let maybe_mnemonic = str_value
+                    .trim()
+                    .split_ascii_whitespace()
+                    .map(|v| v.to_string())
+                    .collect::<Vec<String>>();
```
shekohex, line 68: Would be better to use https://docs.rs/tiny-bip39/latest/bip39/struct.Mnemonic.html#method.from_phrase instead of our custom logic here.

config/development/cosmwasm-localnet/terra.toml
```diff
@@ -0,0 +1,71 @@
...
+# Websocket Endpoint for long living connections
+ws-endpoint = "ws://localhost:26557"
+# chain specific id from cosmwasm opcode
+chain-id = "1234"  # "localterra" fake value
+# # The Private Key of this account on this network
+# # the format is more dynamic here:
+# # 1. if it starts with '0x' then this would be raw (64 bytes) hex encoded
+# #    private key.
+# #    Example: 0x8917174396171783496173419137618235192359106130478137647163400318
+# #
```
shekohex, line 16: Not valid anymore, right? this commit is outdated.

config/development/cosmwasm-localnet/juno.toml
```diff
@@ -0,0 +1,76 @@
...
+# Websocket Endpoint for long living connections
+ws-endpoint = "ws://localhost:26657"
+# chain id
+chain-id = "1233" # "localjuno" fake value
+# # The Private Key of this account on this network
+# # the format is more dynamic here:
+# # 1. if it starts with '0x' then this would be raw (64 bytes) hex encoded
+# #    private key.
+# #    Example: 0x8917174396171783496173419137618235192359106130478137647163400318
+# #
```
shekohex, line 16: Same here, this comment is outdated.

shekohex (changes requested):

config/development/cosmwasm-localnet/juno.toml
```diff
@@ -0,0 +1,76 @@
+# Block which represents properties for a network
+[cosmwasm.juno]
+# The name that the chain is indexed on, for linkable anchors
+name = "localjuno"
+# Http(s) Endpoint for quick Req/Res
+http-endpoint = "http://localhost:26657"
+# Websocket Endpoint for long living connections
+ws-endpoint = "ws://localhost:26657"
+# chain id
+chain-id = "1233" # "localjuno" fake value
```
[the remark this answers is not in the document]
duguorong009: Original plan was to use the real `chain-id` like "localjuno" and calculate the numeric value using the formula we discussed earlier.
Here, I just used the (literally) `fake` value for the development

src/types/mnemonic.rs
duguorong009: One question, @shekohex 
Do you mean we don't introduce the custom type but use the existing `mnemonic` from the `tiny_bip39` crate?

src/types/mnemonic.rs
duguorong009: Ok, will follow your recommend

config/development/cosmwasm-localnet/juno.toml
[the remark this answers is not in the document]
drewstone: Even if we are to put the hardcoded value in there let's use the actual one, not a fake value.

src/types/mnemonic.rs
duguorong009: Done https://github.com/webb-tools/relayer/pull/230/commits/cd60a5f0389f9a322c260705a7a75e7d2aa01095

src/types/mnemonic.rs
duguorong009: Done https://github.com/webb-tools/relayer/pull/230/commits/cd60a5f0389f9a322c260705a7a75e7d2aa01095

config/development/cosmwasm-localnet/juno.toml
duguorong009: Cleaned in https://github.com/webb-tools/relayer/pull/230/commits/5ec0574a1aeb4c16725a460905c3dbbb7df55782

config/development/cosmwasm-localnet/terra.toml
duguorong009: Cleaned in https://github.com/webb-tools/relayer/pull/230/commits/5ec0574a1aeb4c16725a460905c3dbbb7df55782

config/development/cosmwasm-localnet/juno.toml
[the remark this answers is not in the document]
duguorong009: Updated the **fake numeric chain_id**s with **real string one**.
Will add more code for read/convert the string chain_id to numeric one.

src/types/mnemonic.rs
```diff
@@ -0,0 +1,56 @@
...
+                E: serde::de::Error,
+            {
+                BipMnemonic::from_phrase(value, Language::English).map_err(
+                    |_| {
+                        serde::de::Error::custom(format!(
+                            "Cannot get the mnemonic from string: {}",
+                            value
+                        ))
+                    },
+                )
```
shekohex, line 49: That's good, but I guess you forgot and removed the support of the environment variables, the using of $VAR in the config here is important, since these are secrets and it is not wise to leave them hard-coded in config files like this in plain text.

shekohex (changes requested):

src/types/mnemonic.rs
duguorong009: Thanks @shekohex 
Done https://github.com/webb-tools/relayer/pull/230/commits/db5d8f6232b920fa95f1c6cd79af2d7e33d6b728

src/types/mnemonic.rs
```diff
@@ -39,7 +39,30 @@ impl<'de> Deserialize<'de> for Mnemonic {
...
+                    let val = std::env::var(var).map_err(|e| {
+                        serde::de::Error::custom(format!(
+                            "error while loading this env {}: {}",
+                            var, e,
+                        ))
+                    })?;
+                    str_value = val;
+                } else {
+                    str_value = value.to_string();
+                }
```
shekohex, line 64: Here is a better way to write this block:
```suggestion
                let str_value = if value.starts_with("0x") {
                    // hex value
                    return Err(serde::de::Error::custom(format!(
                        "got {} but expected a 12/24 word list ",
                        value
                    )));
                } else if value.starts_with('>') {
                    todo!("Implement command execution to extract the mnemonic")
                } else if value.starts_with('$') {
                    // env
                    let var = value.strip_prefix('$').unwrap_or(value);
                    tracing::trace!("Reading {} from env", var);
                    std::env::var(var).map_err(|e| {
                        serde::de::Error::custom(format!(
                            "error while loading this env {}: {}",
                            var, e,
                        ))
                    })?
                } else {
                    value.to_string()
                };
```

shekohex (changes requested):

PR #236, opened 2022-08-01
Make Secrets Optional

### LINKED ISSUE #234: [TASK] Change `private-key` in the config to be optional, and later be provided during the runtime.
### Context

We need to make the [private_key](https://github.com/webb-tools/relayer/blob/c169b62f560a607535e1155d9f312aa30cfd73a3/src/config.rs#L138), [suri](https://github.com/webb-tools/relayer/blob/c169b62f560a607535e1155d9f312aa30cfd73a3/src/config.rs#L202) and others, optional. The main reason is that we are transitioning to make the relayer as a crate, and we wanted to make providing some information during the runtime, like the private key in this case, instead of the config during parse time. Other reason is that if you are running the relayer as only a data relaying service where you serve the leaves for the dApp, or other applications. in that case, there is no point to define any private key.

### Required Changes

1. Make `private_key`, `suri` ...others `Option<...>`.
2. Add a `verify` method to the `WebbRelayerConfig` to be called during the runtime to verify that the config is correct and can be used, this method should verify each network contains a private key, with conditions like when the relayer needs to use the private key in the future, or ignored at all like when running the relayer as purely data relayer.

shekohex:
## Summary of changes
- See #234 

- Closes #234 

-----
### Code Checklist 

- [ ] Tested
- [ ] Documented

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -102,6 +102,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8da52d66c7071e2e3fa2a1e5c6d088fec47b593032b254f5e980de8ea54454d6"
 
 [[package]]
+name = "async-lock"
+version = "2.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


drewstone (approved):
LGTM

PR #245, opened 2022-08-05
Verify proposal nonce and proposal data before execution

### LINKED ISSUE #208: [BUG] Old SignatureBridge proposal transactions are attempted while syncing
# Current Behavior
If a relayer for a SignatureBridge starts syncing and sees an old proposal, the relayer will still attempt to send a transaction for the proposal:

https://gist.github.com/nepoche/3186788d08081f868fc32d850061e118

# Expected Behavior

The relayer should perform a 'dry-run' similar to the dry-run performed in private transaction relaying flows.

salman01zp:
## Summary of changes
- Verify proposal data. It should be greater than 40B
- Verify proposal nonce. `Proposal nonce` should be greater than Proposal nonce stored at the signature bridge. With this, we can ensure that the same proposal is not executed twice by the relayer.

- Closes #208 

-----
### Code Checklist 

- [ ] Tested
- [ ] Documented

### DIFF
--- a/src/events_watcher/substrate/signature_bridge_watcher.rs
+++ b/src/events_watcher/substrate/signature_bridge_watcher.rs
@@ -131,24 +131,45 @@ where
         chain_id: U256,
         store: Arc<<Self as SubstrateEventWatcher>::Store>,
         api: Arc<<Self as SubstrateEventWatcher>::Api>,
-        (data, signature): (Vec<u8>, Vec<u8>),
+        (proposal_data, signature): (Vec<u8>, Vec<u8>),
     ) -> anyhow::Result<()> {
-        let typed_chain_id =
-            webb_proposals::TypedChainId::Substrate(chain_id.as_u32());
-        let data_hex = hex::encode(&data);
+        let proposal_data_hex = hex::encode(&proposal_data);
+        // 1. Verify proposal length. Proposal lenght should be greater than 40 bytes (proposal header(40B) + proposal body).
+        if proposal_data.len() < 40 {
+            tracing::warn!(
+                proposal_data = ?proposal_data_hex,
+                "Skipping execution of this proposal :  Invalid Proposal",
+            );
+            return Ok(());
+        }
+
+        // 2. Verify proposal nonce. Proposal nonce should be greater than signature bridge proposal nonce.
+        let bridge_proposal_nonce = api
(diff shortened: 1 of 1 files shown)


src/events_watcher/substrate/signature_bridge_watcher.rs
```diff
@@ -279,6 +309,11 @@ where
     }
 }
 
+pub fn parse_nonce_from_proposal_data(proposal_data: &[u8]) -> u32 {
+    let nonce_bytes = proposal_data[36..40].try_into().unwrap_or_default();
+    u32::from_be_bytes(nonce_bytes)
+}
```
shekohex, line 315: I'm sure parsing the full proposal header here is safer, and even more readable:
```suggestion
pub fn parse_nonce_from_proposal_data(proposal_data: &[u8]) -> u32 {
	assert!(proposal_data.len() >= ProposalHeader::LENGTH);
	let mut header_bytes = [0u8; ProposalHeader::LENGTH];
	header_bytes.copy_from_slice(&proposal_data[0..ProposalHeader::LENGTH]);
	let header = ProposalHeader::from(header_bytes);
    header.nonce.to_u32()
}
```

src/events_watcher/substrate/signature_bridge_watcher.rs
```diff
@@ -131,53 +131,83 @@ where
...
         api: Arc<<Self as SubstrateEventWatcher>::Api>,
-        (data, signature): (Vec<u8>, Vec<u8>),
+        (proposal_data, signature): (Vec<u8>, Vec<u8>),
     ) -> anyhow::Result<()> {
-        let typed_chain_id =
-            webb_proposals::TypedChainId::Substrate(chain_id.as_u32());
-        let data_hex = hex::encode(&data);
+        let proposal_data_hex = hex::encode(&proposal_data);
+        // 1. Verify proposal length. Proposal lenght should be greater than 40 bytes (proposal header(40B) + proposal body).
+        if proposal_data.len() < 40 {
```
shekohex, line 138:
```suggestion
        if proposal_data.len() < ProposalHeader::LENGTH {
```

shekohex (changes requested):
Looks good so far! 
I guess you missed to add the same checks for Signature Bridge contract watcher too.

PR #272, opened 2022-09-12
Cross chain transaction

### LINKED ISSUE #240: [SPEC] Allow for Cross-Chain Communication through a new `LinkedAnchor`s format.
# Overview

We want to enable a Cross-Chain communication, it is possible by design from the protocol perspective, but as of now you can not do a Deposit on an EVM chain and then do the withdraw on a Substrate based chain; and this primarily due to that relayers does not have the notion of constructing the cross chain communication, yet! This SPEC is to enable this through a new format for the linked anchors. It contains a summary on the top of how the end goal should be, a more technical details on how we can implement this, and a checklist with tests to ensure everything is working.

## Summary

By the end after implementing this SPEC we should be able to write our configuration for linked anchors a cross all systems in the following way:

```bash
linked-anchors = [
	# Using the raw hex value of the resource id
	# This `ResourceId` contains all the information we need to link to
	# the other chain, it contains the type of the chain, and its Id also
	# the target system, in this case it is the contract address of the anchor
	# handler.
	{ type = "raw", resource-id = "0x000000000000e69a847cd5bc0c9480ada0b339d7f0a8cac2b66701000000138a" }

	# or we can use the more human redable format, for fast prototyping.
	# This format is easier to read by us, it will contain the type of the target
	# chain, the raw id of that chain and lastly, the target system
	# and in this example all we need to know is just the contract address of the
	# anchor handler. (internally, the relayer will construct the correct resource id
	# from these information.)
	{ type = "EVM", chain-id = 5001, address = "0xe69a847cd5bc0c9480ada0b339d7f0a8cac2b667" },

	# Here is another example for a substrate based chain.
	# in this example, we need to know the pallet and the call index (the index
	# of the anchor_handler pallet) and also we should know the tree id.
	{ type = "substrate", chain-id = 5, pallet = 42, call = 2, tree-id = 7 },
	
	# and the cool thing about this is that you can mix and match between
	# these different formats in the same list of the linked anchors.
]
```

Using the above configuration for the linked anchors, the relayer will be able to relay roots to the different systems.

## One Step Deeper

In the relayer `config.rs` We have `LinkedVAnchorConfig` and `SubstrateLinkedVAnchorConfig` but now we will have both of them combined into one enum that looks like this 

```rust
#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
#[serde(tag = "type")]
enum LinkedAnchor {
    // TODO: replace `String` with a type like `ethereum_types::H256`
    Raw { resource_id: String },

    #[serde(rename = "EVM")]
    // TODO: replace `String` with a type like `ethereum_types::Address`
    Evm { chain_id: u32, address: String },
    Substrate {
        chain_id: u32,
        pallet: u8,
        call: u8,
        tree_id: u32,
    },
}
```

See it in action [[Rust playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d3351c273b02c6825b607ba7e337498c)]

The other important thing here to mention is that the relayer will (internally) use the linked anchors as `ResourceId`s and to expand on that, in the config post processing function, we should convert all the linked anchor to just a list of the raw `ResourceId` bytes.

## The Proposal Flow

When we hear a `transcat` happened on one of the `VAnchor` on any chain, the relayer will first look into its linked anchors list, go through each one, look into the target system and the target typed chain id then construct the corresponding `AnchorUpdateProposal` for that system, for example: 

We have Chain A (EVM Based) and Chain B (Substrate Based), then we have Alice here that used our system to do a “Deposit” on Chain A that she wants to do the “Withdraw” on Chain B. Since the relayer know about both, and Chain A is linked to Chain B (and *vice versa*) through the resource Ids. All what the relayer should do in this case to construct the correct `AnchorUpdateProposal` for Substrate Chain B then send it to be signed to the proposal signing backend, after this the proposal should be signed, later the relayer will attempt to send the signed proposal to the signature bridge pallet to be executed. at the end Alice should now be able to do the “Withdraw” on Chain B.

# Checklist

- [x]  Make a spec to change Webb’s `TargetSystem::TreeId` to contain the `pallet` and `call` values.
- [x]  Update EVM’s and Substrate’s config `linked_anchors` to be `Vec<LinkedAnchor>`
- [x]  Modify the `MockedProposalSigningBackend` to have a `Hashset<ResouceId>` instead of `HashMap<TypedChainId, TargetSystem>` for `signature_bridges`.
- [x]  Change what is needed to construct the `MockedProposalSigningBackend` in the `service.rs`
- [x]  For `can_handle_proposal` check, we should just check if the `signature_bridges` set contains the `ResouceId` from the proposal header.
- [x]  For both EVM and Substrate `VAnchor` watchers we should add a method that constructs the `AnchorUpdateProposal` depending on the `ResourceId` of the target system, this function will take the `ResourceId` (from the linked anchors) along with (`merkle_root`, `src_chain_id` and `latest_leaf_index`) and returns one of `substrate::AnchorUpdateProposal` or `evm::AnchorUpdateProposal` …etc.
- [x]  in the `ProposalHandlerWatcher` add support for constructing the correct `BridgeKey` for Substrate based chains. see [here](https://github.com/webb-tools/relayer/blob/ce6ede8bb5da671695856d79e704bfe06086321d/src/events_watcher/dkg/proposal_handler_watcher.rs#L99-L104) for example.
- [x]  Write a test that involves two EVM based chains, one Substrate based chain, and using a Mocked Signing Backend during the this test to do a deposit on one of the EVM chains and the withdraw on the Substrate chain.
- [x]  In the same test above, try the other way around to do a deposit on the Substrate based chain and the Withdraw on the other EVM one.
- [ ]  Write another test similar to the above one, but instead of using the Mocked Signing Backend, use the DKG Signing Backend.
- [ ] Update preset configs to reflect new LinkedAnchors
- [ ] Update documentation in README.md and docs site

salman01zp:
## Summary of changes
-  Cross-chain transaction implementation for the relayer
- Updated linkedAnchor config to support cross-chain configuration

## Note
On vanchor deposit `Relayer` relayer roots to all linked anchors but stuck with the issue on withdrawal flow.
I believe we can merge this PR as all other tests works and the relayer also relays roots correctly for the cross-chain transaction. 

And we can complete tests with withdrawal flow in the next PR @drewstone @shekohex 

- Closes #240 

-----
### Code Checklist 

- [x] Tested
- [x] Documented


shekohex (approved):
I guess this good to be merged now.

PR #284, opened 2022-10-14
wait for n block confirmations for evm chains before processing events

### LINKED ISSUE #193: [TASK] Wait until an event is finalized/N blocks deep before proposing/relaying it.
# Overview
To prevent against the worst-case issue of dealing with forks in a chain, we should ensure that the relayers wait some # of blocks for each different chain before relaying or proposing new updates, especially for anchor updates.

For example, if we relay an anchor update to Ethereum and then Ethereum forks, our updates will be lost. Even worse, if we relay an Ethereum anchor update to its neighbors and Ethereum forks, its neighbors will have a corrupted state of Ethereum. This would be considered a bug in our bridge.

## Spec
In order to prevent against such attacks/issues, we should have a notion of finality for all the chains we are relaying for. If they have finality then we wait explicitly for that (i.e. that an event is in a finalized block) and if it doesn't then we have some length of time we wait before relaying it, such as 10-100 blocks.

We should be able to identify if events have occurred in finalized blocks firstly and we should add this functionality. For L2s connected to Ethereum with no notion of finality, we should identify and implement something that prevents this issue to the best of our ability.

salman01zp:
## Summary of changes
- Add `block_confirmations` parameter in EVM chain config  and use this configuration to get finalized block which we can then process.

```rust
pub struct EvmChainConfig {
    /// String that groups configuration for this chain on a human-readable name.
    pub name: String,
    /// Websocket Endpoint for long living connections
    #[serde(skip_serializing)]
    pub ws_endpoint: RpcUrl,
    /// Block confirmations
    #[serde(skip_serializing)]
    pub block_confirmations: u8,
   .
   .
}

// last processed block
let block = store.get_last_block_number(
       (chain_id, contract.address()),
        contract.deployed_at(),
       )?;
// latest finalized block after n block_confirmations
let finalized_block  = current_block_number.saturating_sub(block_confirmations);
// block we will process
let dest_block = cmp::min(block +1 , finalized_block);
```

## Note

Will add finality config once we get new release form ethers-rs

- Closes #193 

-----
### Code Checklist 

- [x] Tested
- [x] Documented


drewstone:
We should add this for EVM chains that don't have finality. Maybe we can have some notion of this in their config in a parameter like `has_finality = false`? Or we could have a flag that specifies the type of finality as `finality_type = 10` or `finality _type = final` which would indicate:
-  `finality_type = 10` - for soft-finality on the chain (# of confirmations before considered safe)
- `finality _type = final` - for hard finality on the chain (finalized blocks)

salman01zp:
currently, ether-rs have not added a tag for `finalized` blocks. I have created an issue for it [here](https://github.com/gakonst/ethers-rs/issues/1791)

shekohex:
from Drew comment, it could be like the following:

```rust
#[derive(Copy, Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
#[serde(tag = "type")]
enum FinalityConfig {
    Final,
    BlockConfirmations(BlockConfirmationsConfig),
}

#[derive(Copy, Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
struct BlockConfirmationsConfig {
    value: u64,
}
```
So that we can use in the config like the following

```toml
[finality]
type = "final"
```
Or in case of networks that we have to wait some blocks confirmation,

```toml
[finality]
type = "block-confirmations"
value = 128
```

[Rust Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e6067a3aa3792f1ef4e60745c2018788)

shekohex:
> This sounds good to me.
> 
> currently, ether-rs have not added a tag for `finalized` blocks. I have created an issue for it [here](https://github.com/gakonst/ethers-rs/issues/1791) @drewstone @shekohex

We can open a PR to add the new API parameters from this doc [here](https://docs.alchemy.com/reference/ethereum-developer-guide-to-the-merge#what-apis-are-changing) to [this code](https://github.com/gakonst/ethers-rs/blob/a07581489a12b1007c3a261dc8df2bbdc4e27918/ethers-core/src/types/block.rs#L533-L635) in ethers.

salman01zp:
yes, I have opened PR there.

drewstone:
It was added, what's the status of this @salman01zp ?

salman01zp:
@drewstone  will need a new release for ethers including new block tags.

drewstone (approved):
LGTM

tests/lib/localTestnet.ts
```diff
@@ -432,6 +435,7 @@ export class LocalChain {
       'http-endpoint': config.httpEndpoint,
       'ws-endpoint': config.wsEndpoint,
       'chain-id': config.chainId,
+      'block-confirmations': config.blockConfirmations,
       beneficiary: config.beneficiary,
```
drewstone, line 439: Is there a reason some of these have quotes `'` and others don't?

salman01zp: fields with quotes are the ones we skip while serializing  and just convert from `kebab-case`.
```rust
 /// Block confirmations
    #[serde(skip_serializing)]
    pub block_confirmations: u32,
```

we can send it as camelCase field new need rename it like this
```rust
 /// Block confirmations
    #[serde(rename(serialize = "blockConfirmations"))]
    pub block_confirmations: u32,
```

shekohex: Javascript fields names can't have `-` in them, so we need to put them between quotes, but we can convert them as Salman Said to camelCase when we serialize them.

Issue #286, opened 2022-10-19
[TASK] Implement event handler trait for substrate


salman01zp:
Implement the `EventHandler` trait similar to Evm [here](https://github.com/webb-tools/relayer/blob/main/src/events_watcher/mod.rs#L273-#L304) which will allow one event watcher to run and many event handlers that will run in parallel.

```rust

let vanchor_watcher_task = event_watcher.run(
      client,
      store,
      vec![
            Box::new(deposit_handler),
            Box::new(leaves_handler),
            Box::new(encrypted_output_handler),
          ],
       &my_ctx,
       );

```

PR #287, opened 2022-10-19
Encrypted output watcher for substrate

### LINKED ISSUE #276: Store Encrypted Commitments for Substrate
Storing of encrypted commitiments for the EVM is done here #275 

We'd need to do the same for the Substrate protocol. For this to be done, the `VAnchor` pallet in Substrate protocol needs to add the `encrypted_outputs` as part of the events to be emitted. 

Changes also need to be made to the `Webb-rs` repo to accommodate this new event properties to be added to the VAnchor

salman01zp:
## Summary of changes
- [x]  Encrypted output watcher for substrate
- [x]  Removed `ethereum-types` used in relayer storage and substrate watchers
- [x] Updated `HistoryStoreKey` to use ResourceId
## Note 
I have not initialized this watcher since we don't have a substrate event handler with which we can run multiple event handlers for one event watcher. We will initialize it once we have the substrate event handler 
refers to issue #286 
- Closes #276 

-----
### Code Checklist 

- [ ] Tested
- [x] Documented

### DIFF
--- a/src/events_watcher/substrate/vanchor_encrypted_output_watcher.rs
+++ b/src/events_watcher/substrate/vanchor_encrypted_output_watcher.rs
@@ -1,1 +1,100 @@
+// Copyright 2022 Webb Technologies Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+use super::{BlockNumberOf, SubstrateEventWatcher};
+use crate::metric;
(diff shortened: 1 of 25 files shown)


drewstone (approved):
LGTM

src/events_watcher/substrate/vanchor_encrypted_output_watcher.rs
```diff
@@ -1 +1,105 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+use super::{BlockNumberOf, SubstrateEventWatcher};
+use crate::metric;
+use std::sync::Arc;
+use webb::evm::ethers::types;
```
drewstone, line 18: I don't think we should be using this type @salman01zp

drewstone: Since we keep things under separate feature flags too this should not have any EVM depedencies. `u128` is sufficient for the counter.

drewstone (changes requested):

drewstone (approved):

PR #289, opened 2022-10-24
Reorg the entire thang

drewstone:
## Summary of changes
- Moves the system into crates/services/event_watcher libraries.
- Moves services like the webb_relayer into its place.


crates/event-watcher-traits/src/evm/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+use webb_relayer_context::RelayerContext;
+use webb_relayer_store::sled::SledQueueKey;
+use webb_relayer_store::{
+    BridgeCommand, BridgeKey, EventHashStore, HistoryStore, ProposalStore,
+    QueueStore,
+};
+use webb_relayer_utils::metric;
+
+/// Event watching traits
+pub mod event_watcher;
```
shekohex, line 40: make them private mods.
```suggestion
mod event_watcher;
```

shekohex (changes requested):
Here is a graph of the workspace. please be careful with circular dependency. 

graph

crates/beacon-chain-relay/src/test_utils.rs
```diff
@@ -2,7 +2,7 @@ pub fn read_json_file_from_data_dir(file_name: &str) -> std::string::String {
     let mut json_file_path = std::env::current_exe().unwrap();
 
     json_file_path.pop();
-    json_file_path.push("../../../beacon-chain-relay/data");
+    json_file_path.push("../../../crates/beacon-chain-relay/data");
```
salman01zp, line 5:
```suggestion
    json_file_path.push("../../../../crates/beacon-chain-relay/data");
```

PR #306, opened 2022-12-08
Extend relayer metrics

### LINKED ISSUE #305: [SPEC] Extend the Relayer Metrics system
### Overview

Right now the relayer metrics supports tracking for the `gas_spent` and `total_fee_earned` ...etc. But these are too generic, they are calculated over the whole networks and chains. Which is not the best data we shall provide. 

### Solution

Instead of having one total counter for the `gas_spent` and `fees`, we should have many of them in a map between the `ResourceId` and the `Counter`, so that we can query these data across different bridges or different networks ..etc.

We need these information for each Chain/Bridge:
1. Gas Spent.
2. Fees earned.
3. Account balance (as a counter that just have a single value).

salman01zp:
## Summary of changes
- Collect metrics for individual resources

- Closes #305 

-----
### Code Checklist 

- [ ] Tested
- [ ] Documented

### DIFF
--- a/crates/relayer-utils/src/metric.rs
+++ b/crates/relayer-utils/src/metric.rs
@@ -12,10 +12,24 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::collections::HashMap;
+
 use prometheus::core::{AtomicF64, GenericCounter, GenericGauge};
 use prometheus::{register_counter, register_gauge, Encoder, TextEncoder};
+use webb_proposals::ResourceId;
 
-/// A struct definition for collecting metrics in the relayer
+/// A struct for collecting metrics for given resource.
+#[derive(Debug, Clone)]
+pub struct ResourceMetric {
+    /// Total gas spent on Resource.
+    pub total_gas_spent: GenericCounter<AtomicF64>,
+    /// Total fees earned on Resource.
+    pub total_fee_earned: GenericCounter<AtomicF64>,
+    /// Account Balance
+    pub account_balance: GenericGauge<AtomicF64>,
+}
+
+/// A struct definition for collecting metrics in the relayer.
 #[derive(Debug, Clone)]
 pub struct Metrics {
     /// Bridge watcher back off metric
@@ -46,6 +60,8 @@ pub struct Metrics {
     pub gas_spent: GenericCounter<AtomicF64>,
     /// Total amount of data stored metric
     pub total_amount_of_data_stored: GenericGauge<AtomicF64>,
+    /// Resource metric
+    pub resource_metric_map: HashMap<ResourceId, ResourceMetric>,
 }
 
 impl Metrics {
@@ -120,6 +136,9 @@ impl Metrics {
             "The Total number of data stored",
         );
 
+        let resource_metric_map: HashMap<ResourceId, ResourceMetric> =
+            HashMap::new();
+
         Self {
             bridge_watcher_back_off: bridge_watcher_back_off_counter.unwrap(),
             total_active_relayer: total_active_relayer_counter.unwrap(),
@@ -142,6 +161,7 @@ impl Metrics {
             gas_spent: gas_spent_counter.unwrap(),
             total_amount_of_data_stored: total_amount_of_data_stored_counter
                 .unwrap(),
+            resource_metric_map,
         }
     }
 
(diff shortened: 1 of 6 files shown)


crates/tx-queue/src/evm/evm_tx_queue.rs
```diff
@@ -300,6 +299,9 @@ where
                             );
                         }
                     };
+
+                    // gas spent metric
+                    metrics.gas_spent.inc_by(gas_price.as_u64() as f64);
```
shekohex, line 304: This not the correct way to do it, since this does not reflect the real `gas_spent`, for that, after the transaction is mined, you can get these information from the Receipt See https://docs.rs/ethers/latest/ethers/types/struct.TransactionReceipt.html#structfield.gas_used

crates/tx-relay/src/evm/vanchor.rs
```diff
@@ -188,9 +189,21 @@ pub async fn handle_vanchor_relay_tx<'a>(
...
+    handle_evm_tx(call, stream, cmd.chain_id).await;
+
+    // update metric for total fee earned for given resource
+    let target_system = TargetSystem::new_contract_address(
+        contract_config.common.address.to_fixed_bytes(),
+    );
+    let typed_chain_id = TypedChainId::Evm(chain.chain_id);
+    let resource_id = ResourceId::new(target_system, typed_chain_id);
+    let resource_metric =
+        ctx.metrics.resource_metric_map.get(&resource_id).unwrap();
```
shekohex, line 201: Instead of unwrapping, try use [`HashMap::entry`](https://doc.rust-lang.org/stable/std/collections/hash_map/struct.HashMap.html#method.entry) and https://doc.rust-lang.org/stable/std/collections/hash_map/enum.Entry.html#method.or_insert_with if the value is not there.

crates/tx-relay/src/substrate/vanchor.rs
```diff
@@ -108,4 +111,24 @@ pub async fn handle_substrate_vanchor_relay_tx<'a>(
...
+    };
+    let target = SubstrateTargetSystem::builder()
+        .pallet_index(pallet_index)
+        .tree_id(cmd.id)
+        .build();
+    let target_system = TargetSystem::Substrate(target);
+    let typed_chain_id = TypedChainId::Substrate(cmd.chain_id as u32);
+    let resource_id = ResourceId::new(target_system, typed_chain_id);
+    let resource_metric =
+        ctx.metrics.resource_metric_map.get(&resource_id).unwrap();
```
shekohex, line 128: Same here, as above, please do not use unwrap.

shekohex (changes requested):

crates/relayer-handlers/src/routes/metric.rs
```diff
@@ -0,0 +1,136 @@
...
+/// Handles relayer metric requests for evm based resource
+///
+/// Returns a Result with the `ResourceMetricResponse` on success
+pub async fn handle_evm_metric_info(
+    chain_id: u32,
+    contract: Address,
+    ctx: Arc<RelayerContext>,
+) -> Result<impl warp::Reply, Infallible> {
+    let metrics_clone = ctx.metrics.clone();
+    let mut metrics = match metrics_clone.try_lock() {
```
shekohex, line 51: Question: Why here we are using `try_lock` instead of just `lock().await`?

crates/relayer-handlers/src/routes/metric.rs
```diff
@@ -0,0 +1,136 @@
...
+/// Returns a Result with the `ResourceMetricResponse` on success
+pub async fn handle_substrate_metric_info(
+    chain_id: u32,
+    tree_id: u32,
+    pallet_id: u8,
+    ctx: Arc<RelayerContext>,
+) -> Result<impl warp::Reply, Infallible> {
+    let metrics_clone = ctx.metrics.clone();
+    // let mut metrics = metrics_clone.try_lock();
+    let mut metrics = match metrics_clone.try_lock() {
```
shekohex, line 98: Same here, why not just `metrics_clone.lock().await`

crates/relayer-utils/src/metric.rs
```diff
@@ -120,6 +136,9 @@ impl Metrics {
             "The Total number of data stored",
         );
 
+        let resource_metric_map: HashMap<ResourceId, ResourceMetric> =
+            HashMap::new();
+
```
shekohex, line 141: No types are needed here, rustc can infer that.

```suggestion
        let resource_metric_map = HashMap::new();
```

crates/tx-relay/src/substrate/vanchor.rs
```diff
@@ -108,4 +112,33 @@ pub async fn handle_substrate_vanchor_relay_tx<'a>(
     };
 
     handle_substrate_tx(event_stream, stream, cmd.chain_id).await;
+
+    let pallet_index = {
+        let metadata = client.metadata();
+        let pallet = metadata.pallet("VAnchorHandlerBn254").unwrap();
```
shekohex, line 118: That's interesting, `unwrap` here is not something we want for sure, what we can do instead is return the `pallet_index` as an Option, where if the pallet `VAnchorHandlerBn254` does not exist in the runtime we return `None`, otherwise we resolve the pallet index. This way, in the next line we match over the pallet_index, and if it is `None` we return early from the function.

tests/lib/webbRelayer.ts
```diff
@@ -559,6 +565,12 @@ export interface RelayerMetricResponse {
   metrics: string;
 }
 
+export interface ResourceMetricResponse {
+  totalGasSpent: number,
+  totalFeeEarned: number,
+  accountBalance: number
+}
```
shekohex, line 572: These are strings, no?
```suggestion
export interface ResourceMetricResponse {
  totalGasSpent: string,
  totalFeeEarned: string,
  accountBalance: string
}
```

shekohex (changes requested):

services/webb-relayer/src/main.rs
```diff
@@ -53,18 +53,20 @@ async fn main(args: Opts) -> anyhow::Result<()> {
...
     // metric for data stored which is determined every 1 hour
     let sled_metric_task_handle = tokio::task::spawn(async move {
         let mut sled_data_metric_interval =
             time::interval(Duration::from_secs(3600));
         loop {
             sled_data_metric_interval.tick().await;
             // set data stored
-            cloned_ctx
-                .metrics
+            let metrics = metrics_clone.lock().await;
```
tbraun96, line 65: Converge into one line

tbraun96 (changes requested):
Looks mostly good, just some language features that we can take advantage of the reduce the number of lines of code.

crates/relayer-handlers/src/routes/metric.rs
salman01zp: Yes, Initially I wanted to handle the case if the lock was unavailable by using `try_lock` and avoid blocking resources. But on second thought it will be better to use `lock.await` and process the requests, since there won't be that many requests which will slow down the relayer.

drewstone:
Fixed a merge conflict here @salman01zp, if tests fail please make sure to check on it.

PR #322, opened 2023-01-09
Remove unused dependencies

Nutomic:
## Summary of changes
- Remove unused dependencies according to [cargo-machete](https://lib.rs/crates/cargo-machete) and [cargo-udeps](https://lib.rs/crates/cargo-udeps). This doesnt decrease the number of compiled unites, but reduces clean build time from 6m 4s to 5m 5s, most likely due to increased parallelism.
- Disable debug info and strip symbols. This means debuggers and stack traces dont work anymore, but it significantly decreases build time to 1m 59s (unless i measured wrong).

By the way, the `native-tls` dependency which is listed in most crates also seems to be unused. What do you think about moving it to [webb-rs](https://github.com/webb-tools/webb-rs/blob/main/Cargo.toml)? Would make more sense to me.

If this looks good then I can apply the same changes to the other Rust repos.

- Closes 

-----
### Code Checklist 

- [ ] Tested
- [ ] Documented


drewstone:
Great work @Nutomic !

PR #325, opened 2023-01-11
Implement fee estimate

Nutomic:
Very much work in progress

## Summary of changes
- Add API endpoint to calculate fees
- Estimate transaction fee using etherscan.io API
- Calculate token exchange rate using coingecko.com

## Todo
- [x] Actually use the calculated fee in relay
- [ ] Support for different blockchains
- [ ] Retrieve actual token names
- [ ] Error handling
- [ ] Fix todos
- [ ] Update API description in readme

- Closes 

-----
### Code Checklist 

- [x] Tested
- [ ] Documented

### DIFF
--- a/crates/relayer-handlers/src/lib.rs
+++ b/crates/relayer-handlers/src/lib.rs
@@ -217,13 +217,19 @@ pub async fn handle_substrate<'a>(
     }
 }
 
-// TODO: do we need to add another endpoint for substrate?
+/// Handler for fee estimation
+///
+/// # Arguments
+///
+/// * `chain_id` - An U256 representing the chain id of the chain to query
+/// * `ctx` - RelayContext reference that holds the configuration
+///
+/// TODO: do we need to add another endpoint for substrate?
 pub async fn handle_fee_info(
     chain_id: u32,
-    client_ip: IpAddr,
     ctx: Arc<RelayerContext>,
 ) -> Result<impl warp::Reply, Infallible> {
-    let fee_info = calculate_fees(chain_id, client_ip, ctx).await.unwrap();
+    let fee_info = calculate_fees(chain_id, ctx).await.unwrap();
 
     Ok(warp::reply::with_status(
         warp::reply::json(&fee_info),
--- a/crates/relayer-handlers/src/lib.rs
+++ b/crates/relayer-handlers/src/lib.rs
@@ -19,7 +19,6 @@
 use std::convert::Infallible;
 use std::error::Error;
 use std::net::{IpAddr, SocketAddr};
-use std::sync::Arc;
 
 use futures::prelude::*;
 
@@ -219,17 +218,9 @@ pub async fn handle_substrate<'a>(
 
 /// Handler for fee estimation
 ///
-/// # Arguments
-///
-/// * `chain_id` - An U256 representing the chain id of the chain to query
-/// * `ctx` - RelayContext reference that holds the configuration
-///
 /// TODO: do we need to add another endpoint for substrate?
-pub async fn handle_fee_info(
-    chain_id: u32,
-    ctx: Arc<RelayerContext>,
-) -> Result<impl warp::Reply, Infallible> {
-    let fee_info = calculate_fees(chain_id, ctx).await.unwrap();
+pub async fn handle_fee_info() -> Result<impl warp::Reply, Infallible> {
+    let fee_info = calculate_fees().await.unwrap();
 
     Ok(warp::reply::with_status(
         warp::reply::json(&fee_info),
--- a/crates/relayer-handlers/src/lib.rs
+++ b/crates/relayer-handlers/src/lib.rs
@@ -19,6 +19,7 @@
 use std::convert::Infallible;
 use std::error::Error;
 use std::net::{IpAddr, SocketAddr};
+use std::sync::Arc;
 
 use futures::prelude::*;
 
@@ -32,6 +33,7 @@ use webb_relayer_handler_utils::{
     IpInformationResponse, SubstrateCommand,
 };
 
+use crate::routes::fees::calculate_fees;
 use webb_relayer_tx_relay::evm::vanchor::handle_vanchor_relay_tx;
 use webb_relayer_tx_relay::substrate::mixer::handle_substrate_mixer_relay_tx;
 use webb_relayer_tx_relay::substrate::vanchor::handle_substrate_vanchor_relay_tx;
@@ -213,4 +215,18 @@ pub async fn handle_substrate<'a>(
             handle_substrate_vanchor_relay_tx(ctx, cmd, stream).await;
         }
     }
+}
+
+// TODO: do we need to add another endpoint for substrate?
+pub async fn handle_fee_info(
+    chain_id: u32,
+    client_ip: IpAddr,
+    ctx: Arc<RelayerContext>,
+) -> Result<impl warp::Reply, Infallible> {
+    let fee_info = calculate_fees(chain_id, client_ip, ctx).await.unwrap();
+
+    Ok(warp::reply::with_status(
+        warp::reply::json(&fee_info),
+        warp::http::StatusCode::OK,
+    ))
 }
--- a/crates/relayer-handlers/src/lib.rs
+++ b/crates/relayer-handlers/src/lib.rs
@@ -217,13 +217,17 @@ pub async fn handle_substrate<'a>(
 }
 
 /// Handler for fee estimation
-///
-/// TODO: do we need to add another endpoint for substrate?
 pub async fn handle_fee_info() -> Result<impl warp::Reply, Infallible> {
-    let fee_info = calculate_fees().await.unwrap();
-
-    Ok(warp::reply::with_status(
-        warp::reply::json(&fee_info),
-        warp::http::StatusCode::OK,
-    ))
+    let fee_info = calculate_fees()
+        .await
+        .and_then(|f| Ok(serde_json::to_string(&f)?));
+    Ok(match fee_info {
+        Ok(value) => {
+            warp::reply::with_status(value, warp::http::StatusCode::OK)
+        }
+        Err(e) => warp::reply::with_status(
+            e.to_string(),
+            warp::http::StatusCode::INTERNAL_SERVER_ERROR,
+        ),
+    })
 }
--- a/crates/relayer-handlers/src/routes/fees.rs
+++ b/crates/relayer-handlers/src/routes/fees.rs
@@ -9,7 +9,5 @@ pub async fn calculate_fees() -> webb_relayer_utils::Result<FeeInfo> {
     let wrapped_token = "usd-coin";
     let base_token = "ethereum";
 
-    let fee_info =
-        get_fee_info(estimated_gas_amount, wrapped_token, base_token).await;
-    Ok(fee_info)
+    get_fee_info(estimated_gas_amount, wrapped_token, base_token).await
 }
--- a/crates/relayer-handlers/src/routes/fees.rs
+++ b/crates/relayer-handlers/src/routes/fees.rs
@@ -1,105 +1,44 @@
-use chrono::DateTime;
-use chrono::Duration;
-use chrono::Utc;
-use coingecko::CoinGeckoClient;
 use serde::Serialize;
-use std::net::IpAddr;
 use std::sync::Arc;
-use std::sync::Mutex;
-use webb::evm::ethers;
-use webb::evm::ethers::types::Chain;
 use webb_relayer_context::RelayerContext;
+use webb_relayer_utils::fees::{calculate_exchange_rate, estimate_gas_price};
 
-static FEE_DATA: Mutex<Vec<FeeInfo>> = Mutex::new(Vec::new());
-
-#[derive(Debug, Serialize, Clone)]
+/// Return value of fee_info API call. Contains information about relay transaction fee and refunds.
+#[derive(Debug, Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct FeeInfo {
-    timestamp: DateTime<Utc>,
-    client_ip: IpAddr,
+    /// Estimated fee for an average relay transaction, in `wrappedToken`
+    /// TODO: We dont know the actual gas amount of the transaction here, so it would make more
+    ///       sense to return a gas price (or multiple), and then calculate the fee on the client.
     fee: f32,
+    /// Exchange rate for refund from `wrappedToken` to `nativeToken`
     refund_exchange_rate: f32,
+    /// Maximum amount of `wrappedToken` which can be exchanged to `nativeToken` by relay
     max_refund: f32,
 }
 
+/// Calculate fee for an average transaction over the relay. Also returns information about refund.
 pub async fn calculate_fees(
     chain_id: u32,
-    client_ip: IpAddr,
     ctx: Arc<RelayerContext>,
 ) -> webb_relayer_utils::Result<FeeInfo> {
     let chain = ctx.config.evm.get(&chain_id.to_string()).unwrap();
 
     // TODO: these are all hardcoded for now
     let gas_amount = 1_721_713;
-    let max_refund = 10.;
+    let max_refund_native = 1.;
 
     let gas_price = estimate_gas_price().await.unwrap();
-    dbg!(gas_price);
 
     let exchange_rate =
         calculate_exchange_rate("usd-coin", "ethereum").await as f32;
-    dbg!(exchange_rate);
 
-    // TODO: we might want to multiply this by a constant factor, in case the actual gas price
-    //       goes up by the time the transaction is submitted
     let native_fee = (gas_price * gas_amount) as f32;
     let wrapped_fee = native_fee * exchange_rate;
-    clear_outdated_fee_data();
+    let max_refund_wrapped = max_refund_native * exchange_rate;
     Ok(FeeInfo {
-        timestamp: Utc::now(),
-        client_ip,
         fee: wrapped_fee,
         refund_exchange_rate: exchange_rate,
-        max_refund,
+        max_refund: max_refund_wrapped,
     })
-}
-
-/// Relay fees are only valid for a limited amount of time. When that time is expired, remove
-/// the entry so that it can't be used anymore.
-fn clear_outdated_fee_data() {
-    let mut lock = FEE_DATA.lock().unwrap();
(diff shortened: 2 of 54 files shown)


Nutomic:
Updated. The basic functionality is working according to the spec as far as I can tell. However there are a lot of todos where things are unclear, please have a look.

The main task now is to retrieve the actual `wrappedToken` and `baseToken` from the transaction. I looked through the code but couldnt find anything related. Any hints how to do this?

crates/tx-relay-utils/src/lib.rs
```diff
@@ -99,7 +99,7 @@ pub struct ExtData<E, I, B, A, T> {
     pub ext_amount: A,
     /// Fee to pay the relayer
     pub fee: B,
-    /// Refund fee
+    /// Refund amount
```
Nutomic, line 102: Seems like a typo.

dutterbutter:
> Updated. The basic functionality is working according to the spec as far as I can tell. However there are a lot of todos where things are unclear, please have a look.
> 
> The main task now is to retrieve the actual `wrappedToken` and `baseToken` from the transaction. I looked through the code but couldnt find anything related. Any hints how to do this?

@shekohex @drewstone any suggestions?

crates/relayer-handlers/src/routes/fees.rs
```diff
@@ -0,0 +1,40 @@
...
+use webb_relayer_utils::fees::{
+    calculate_exchange_rate, calculate_wrapped_fee,
+};
+
+/// Return value of fee_info API call. Contains information about relay transaction fee and refunds.
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct FeeInfo {
+    /// Estimated fee for an average relay transaction, in `wrappedToken`
+    /// TODO: We dont know the actual gas amount of the transaction here, so it would make more
```
drewstone, line 13: We should start initially with a hardcoded gas amount for a given transaction. For example, a transaction with 2 or 16 inputs. I think we could use a `struct FeeConfig` that contains default values set to our "latest" benchmarks.

Over time, as relayers make transactions, they can track the time-series of their gas amounts/prices and eventually average or choose custom strategies here.

drewstone: Worth re-emphasizing since its probably new but transactions also can be different, specifically in the number of inputs they are spending. For now a single `calculate_fees` is sufficient but we will want to take into account that a transaction spending 16 inputs will likely cost a slight bit more gas than one with 2 inputs. This API should grow to reflect that too.

crates/relayer-handlers/src/routes/fees.rs
```diff
@@ -0,0 +1,40 @@
...
+    /// Maximum amount of `wrappedToken` which can be exchanged to `nativeToken` by relay
+    max_refund: U256,
+}
+
+/// Calculate fee for an average transaction over the relay. Also returns information about refund.
+pub async fn calculate_fees() -> webb_relayer_utils::Result<FeeInfo> {
+    // TODO: hardcoded
+    let estimated_gas_amount = U256::from(1_721_713);
+
+    // TODO: need to get the actual tokens which are being exchanged (also in handle_vanchor_relay_tx)
```
drewstone, line 28: Correct, currently each "bridge" or "set of anchors" is directly coupled with a specific Webb wrapped token. If a relayer receives a request to submit a transaction against **Anchor A on Ethereum** then we should be able to directly check **A.token()**.

drewstone: You can organize as you think is best, perhaps with function args or new structs.

crates/relayer-utils/src/fees.rs
```diff
@@ -0,0 +1,84 @@
...
+/// Number of digits after the comma of USD-Coin.
+const USDC_DECIMALS: u32 = 6;
+
+static COIN_GECKO_CLIENT: Lazy<CoinGeckoClient> =
+    Lazy::new(CoinGeckoClient::default);
+
+static ETHERSCAN_CLIENT: Lazy<etherscan::Client> = Lazy::new(|| {
+    etherscan::Client::builder()
+        // TODO: Need to add actual api key via config to increase rate limit
+        .with_api_key("YourApiKeyToken")
```
drewstone, line 19:
```suggestion
        .with_api_key("ETHERSCAN_API_KEY")
```

crates/relayer-utils/src/fees.rs
```diff
@@ -0,0 +1,84 @@
...
+    to_u256(wrapped_fee)
+}
+
+/// Pull USD prices of wrapped token and base token from coingecko.com, and use these to
+/// calculate the exchange rate.
+pub async fn calculate_exchange_rate(
+    wrapped_token: &str,
+    base_token: &str,
+) -> f64 {
+    let tokens = &[wrapped_token, base_token];
```
drewstone, line 43: At launch the `wrapped_token` is likely not going to be listed on any exchanges or markets and therefore no prices. The division at the end seems problematic. Let's just use the `base_token` here.

crates/relayer-utils/src/fees.rs
```diff
@@ -0,0 +1,84 @@
...
+        .await
+        .unwrap();
+    let wrapped_price = prices[wrapped_token].usd.unwrap();
+    let base_price = prices[base_token].usd.unwrap();
+    wrapped_price / base_price
+}
+
+/// Estimate gas price using etherscan.io. Note that this functionality is only available
+/// on mainnet.
+async fn estimate_gas_price() -> crate::Result<u64> {
```
drewstone, line 55: To your previous question about support for different blockchains, here is a perfect function (as many of the others also are) to parameterize this by the chain. Currently, we have the notion of a `TypedChainId` that we use to identify chains and ecosystems. You can find this in [webb-rs](https://github.com/webb-tools/webb-rs) and in Typescript.

The idea with our `TypedChainId` is that it identifies a blockchain ecosystem and a chain in that ecosystem. We needed a unique identifier that could work for different ecosystems where chains shared the same chain identifier.  It fits into a `u48` (or 6-bytes), but we use `u64` more commonly and reserve 8-bytes for this identifier in our protocol messages. An example of an 8-byte (or 6-byte.. drop the 0x0000) `TypedChainId` is 0x0000010000000001. The last 4-bytes represent the `u32` valued chain identifier of all EVM and Substrate based chains.

We could match on different ecosystems/chain identifiers and switch where and how we query for gas. The only two relevant ecosystems atm are EVMs and Substrate chains.

drewstone: For future consideration, we have in the past tried to integrate this protocol with CosmWasm (Cosmos based smart contract system) chains. In Cosmos, people use very long chain identifiers. So while we have this `u64` based identifier which supports nearly every ecosystem, it does not support Cosmos chains.

Example: https://github.com/cosmos/chain-registry/blob/master/gravitybridge/chain.json#L8 (I think this exceeds a `u64`)

I suppose as we start to consider v2 of the protocol we really do need to ensure our protocol messages support the variation in chain identifiers, addresses, etc.

crates/relayer-utils/src/fees.rs
```diff
@@ -0,0 +1,84 @@
...
+    to_u256(max_refund_wrapped)
+}
+
+/// To match types of `ExtData.refund` and `ExtData.fee`, methods here need to return U256,
+/// meaning a conversion is necessary. This conversion is done analogous to
+/// `ethers::utils::parse_ether`, with the actual amount of digits of USDC.
+///
+/// TODO: Needs to support other wrapped tokens with different number of digits. It would be easier
+///       to simply return the f64 amount, but that would require changing the type param for `ExtData`.
+fn to_u256(amount: f64) -> U256 {
```
drewstone, line 80: I'm hesitant to use this data type. It's better to get comfortable with large integer arithmetic IMO in the long run, given all these blockchains deal with `BigInteger` like structures. Another thing to consider is that most of these blockchains use a base unit that has 18 decimals, so 1000 tokens is really 1000 * 10^18.

crates/tx-relay/src/evm/vanchor.rs
```diff
@@ -198,6 +181,35 @@ pub async fn handle_vanchor_relay_tx<'a>(
...
+        withdraw_config.withdraw_fee_percentage,
+        cmd.ext_data.ext_amount.0.abs().as_u128().into(),
+    );
+    */
+    // TODO: Just like refund, this check can fail unexpectedly if exchange rate or gas price change.
+    //       Its probably better to lock in the price for each user at the time fee_info is called,
+    //       and use the exact same exchange rate/transaction fee here.
+    //       This could be done based on IP address, but its fragile because multiple users can have
+    //       the same IP. Better to return a token with `FeeInfo` which then gets passed to
+    //       `handle_vanchor_relay_tx()` in order to use that preagreed fee (for a limited time).
```
drewstone, line 202: Alternatively, we can have the relayers work initially where they lock in a base fee for say 1 minute. User's prove using that fee. If the relayer changes it at the 1 (2, 3, etc.) minute mark and decides that the tx falls out of its threshold (whatever that might mean atm) then it just sends back a request saying the user needs to re-submit with this window's fee.

crates/tx-relay/src/evm/vanchor.rs
```diff
@@ -198,6 +181,35 @@ pub async fn handle_vanchor_relay_tx<'a>(
...
+    /*
+    let expected_fee = calculate_fee(
+        withdraw_config.withdraw_fee_percentage,
+        cmd.ext_data.ext_amount.0.abs().as_u128().into(),
+    );
+    */
+    // TODO: Just like refund, this check can fail unexpectedly if exchange rate or gas price change.
+    //       Its probably better to lock in the price for each user at the time fee_info is called,
+    //       and use the exact same exchange rate/transaction fee here.
+    //       This could be done based on IP address, but its fragile because multiple users can have
```
drewstone, line 200: I like this idea in theory but it sounds like a lot of additional processing we don't need to consider at the moment.

crates/relayer-handlers/src/lib.rs
```diff
@@ -214,3 +215,15 @@ pub async fn handle_substrate<'a>(
         }
     }
 }
+
+/// Handler for fee estimation
+///
+/// TODO: do we need to add another endpoint for substrate?
+pub async fn handle_fee_info() -> Result<impl warp::Reply, Infallible> {
```
drewstone, line 222: Coming back to this now at the end of my review after understanding the flows more..

I think we need to parameterize these functions or wrap them in some parameterizable struct by the
- `TypedChainId`
- `Transaction` type (2 input transfer, 16 input transfer, 2 input withdrawal, 16 input withdrawal)
- `VAnchor` address in question, after all the honest user is requesting info since they're about _to relay a transaction spending some funds whose destination chain is the `TypedChainId`_.

This will tell us the type of chain under consideration so we can switch APIs and gas estimation issues.

drewstone (changes requested):
Good start, hopefully my answers help give more direction here. Let me know if some things aren't clear.

drewstone:
For dealing with the Webb wrapped token in case I didn't add enough details, I would recommend checking out the API for submitting a relayable transaction. In that logic, there should be a clear `VAnchor` contract address that is being called. On that contract there's a getter `function token(): address`. If we can grab an instance of the `VAnchor` contract then we can get its token and from there its base tokens, etc.

crates/relayer-utils/src/fees.rs
Nutomic: In case of USDC there are 6 digits, maybe its a special case.

https://etherscan.io/token/0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48

Nutomic:
Im currently working through the comments and added the caching of `FeeInfo` (single instance, valid for one minute.

I also added fee info to the test case `should relay private transaction`. After adding the fee to `signatureVBridge.transact()` call, it throws an error [UNPREDICTABLE_GAS_LIMIT](https://docs.ethers.org/v5/troubleshooting/errors/#help-UNPREDICTABLE_GAS_LIMIT). The suggested solution is to manually set a gas limit, but `transact()` has no such method. So its probably necessary to add this param in `webb-tools` library, or maybe there is another solution.

crates/relayer-utils/src/fees.rs
Nutomic: In Rust the normal solution for such different data types is something like:
```
enum ChainTypes {
    Ethereum(u64),
    Cosmos(String),
}
```

Serde has [a couple of different ways](https://serde.rs/enum-representations.html) to serialize this

crates/relayer-utils/src/fees.rs
Nutomic: And the main reason this conversion is needed is because coingecko api returns token exchange rates in f64. So its not clear to me how that could be correctly represented as U256.

crates/tx-relay/src/evm/vanchor.rs
Nutomic: Removed it for now.

crates/tx-relay/src/evm/vanchor.rs
Nutomic: This is implemented now with 1 minute lock time for the fee.

crates/relayer-handlers/src/routes/fees.rs
Nutomic: I think you mean `VAnchorContract.token()`. That only returns the contract address, so apparently its still necessary to add a mapping `contract address -> token name`. On the Typescript side in tests there are `name` and `symbol` fields for tokens but these dont seem to be available on the Rust side.

Nutomic:
> For dealing with the Webb wrapped token in case I didn't add enough details, I would recommend checking out the API for submitting a relayable transaction. In that logic, there should be a clear VAnchor contract address that is being called. On that contract there's a getter function token(): address. If we can grab an instance of the VAnchor contract then we can get its token and from there its base tokens, etc.

Im trying this in `handle_vanchor_relay_tx()` and got the `VAnchor.token()` address. However I dont see how to extract the base token from that. Going through etherscan isnt a solution because it wont work for local testnets. So can you give me a hint for this part?

crates/relayer-utils/src/fees.rs
shekohex: Remember we need a binary format that so small and so efficient, easy to implement and test. Not just that, but also it should be fairly easy to implement in different languages and platforms.

Nutomic: In that case we could take a hash of the chain id string, and store the first 64 bits of that hash. That would be enough to check that the chain id matches another id.

Nutomic:
Salman suggested to merge this PR first and finish the remaining tasks in a separate PR, so that its easier to review. We can do this if you want, but the code is still not fully working, except in a very specific scenario. I will list the main changes and remaining tasks here:

Main changes:
- Added fee_info API endpoint `/api/v1/fee_info`:
  ```json
  {
      "estimatedFee": "0x476b26e0f",
      "gasPrice": "0x11",
      "refundExchangeRate": "0x28f",
      "maxRefund": "0xf3e59",
      "timestamp": "2023-01-19T06:29:49.556114073Z"
  }
  ```
- Removed WithdrawConfig which is now unused
- Updated test `should relay private transaction` to use fee info
- Tests now use `cargo build` instead of directly executing binary in target folder, so it is automatically recompiled if anything changed in the Rust code

Todo now:
- Relayer should make a profit, but this is currently not the case. Probably need to increase exchange rate by 1% or something similar.
- fee_info API endpoint doesnt support different chains or tokens yet
- Relaying currently tends to fail fee check, because it estimates gas amount in a different way which usually higher than the hardcoded gas amount used in fee_info API endpoint. Need to decide how to handle this.
- There is no error handling, everything is unwrapped

Todo potentially in separate PRs:
- Token names are still hardcoded to USDC and Ethereum, need to get actual tokens (this could be done in separete PR)
- Checking the refund amount in test "should relay private transaction" is currently not working (Salman said he would debug this)
- fee info needs param for transaction type (number of inputs/outputs), and use different gas amount estimate accordingly
- fee info is cached globally, should be separate cache entries per user/token/chain
- Need to add fee estimation endpoint for substrate

Nutomic:
This is ready to review. After review is done, I we can keep the PR open and wait until I finish the remaining tasks in a second PR, then merge both of them together. This way we avoid breaking the main branch, as the code currently only handles fees for relaying USDC on Ethereum.

shekohex (approved):
LGTM! Good job @Nutomic 

Looking forward for the new PRs!

crates/relayer-utils/src/fees.rs
drewstone: We could do hashing yes, in that case it might fit into our currently allotted 6 bytes. This is actually what we had already planned for the cosmos work.

Nonetheless, the main focus initially is support for EVM and Substrate chains (where we have our current implementations). I would also please ensure you try and leverage the tools we already have built rather than create new ones, [specifically the `TypedChainId`s for identifying chains](https://github.com/webb-tools/webb-rs/blob/main/proposals/src/header.rs#L49-L70).

drewstone:
This looks great too! Great work @Nutomic

crates/relayer-utils/src/fees.rs
Nutomic: I mentioned that in my other pull request:

> I tried to use TypedChainId but thats not possible because Typescript tests use randomly generated chain ids, and converting them to TypedChainId fails. At the same time I noticed that there are many different types used for chain id (string, u32, u64). It would be good to refactor this in the future for more consistency.

drewstone:
We aren’t actually dealing with any string chain ids at the moment too.
Cosmos was just mentioned for the future consideration. We can hash to u32
as well in that case.

drewstone:
Can you say more about the errors using the TyoedChainId in Typescript? I’m
pretty sure we have examples using it.

Nutomic:
@drewstone TypedChainId doesnt support the chain id which is used in Typescript tests which is 3334. Calling `TypedChainId::from(3334)` returns None. So passing the id this way doesnt work.

drewstone:
@Nutomic please refer to: https://github.com/webb-tools/webb-rs/blob/main/proposals/src/header.rs. 

There is a method to get it from a u64. But really I want you to read how we construct it from bytes as that is the encoding that we want. As far as you've mentioned, 3334 fits within a u32. If it's an EVM chain the prefixed 2 bytes should reflect that.

Nutomic:
Closing this one, we can merge #330 only.

PR #327, opened 2023-01-12
Add Support for Pagination while Query the Leaves/EncryptedOutput from the Relayer API

shekohex:
## Summary of changes
- [x] Adds Pagination Support at the Database level for optimized query.
- [x] Adds Support for the Pagination for the API endpoints
	- [x] Leaves (EVM & Substrate)
	- [x] EncryptedOutputs 

### Examples

* Without any range provided (i.e query all the leaves):
 
```
/api/v1/encrypted_outputs/evm/3333/0x1f7a0131A7D39BE4057be8E1006E8E6AdC8C7e50
```

* With Start and End for query a range:

```
/api/v1/encrypted_outputs/evm/3333/0x1f7a0131A7D39BE4057be8E1006E8E6AdC8C7e50?start=20&end=30
```

* With just the start (end is the `MAX` by default)

``` 
/api/v1/encrypted_outputs/evm/3333/0x1f7a0131A7D39BE4057be8E1006E8E6AdC8C7e50?start=50
```

* With just the end (start is `ZERO` by default)

``` 
/api/v1/encrypted_outputs/evm/3333/0x1f7a0131A7D39BE4057be8E1006E8E6AdC8C7e50?end=50
```

For all the above examples, you can replace `encrypted_outputs` with `leaves` and it should work.

-----
### Code Checklist 

- [x] Tested
- [x] Documented


crates/relayer-handlers/src/routes/mod.rs
```diff
@@ -18,3 +18,51 @@ pub mod info;
...
+    }
+}
+
+impl From<OptionalRangeQuery> for core::ops::Range<u32> {
+    fn from(range: OptionalRangeQuery) -> Self {
+        let start = range
+            .start
+            .or_else(default_zero)
+            .expect("start is not None");
+        let end = range.end.or_else(default_u32_max).expect("end is not None");
```
salman01zp, line 57: should we define the some pagination limit for API 
`end = start + pagination_limit.`

shekohex: `u32::MAX` is the default, since the max leaves in one tree ever is `2^depth - 1` and the `depth` for us mostly is `30`.

shekohex: So we are above the max, which is okay.

drewstone (approved):
LGTM

salman01zp (approved):
LGTM..! Great work @shekohex

PR #332, opened 2023-01-30
Improved error handling and code cleanup in tx-relay crate

Nutomic:
## Summary of changes
- Clean up error handling in tx-relay to use Result and `?` operator. This simplifies control flow and makes the code much more readable. However its far from perfect as the "error" type used is `CommandResponse` which is also used for success messages. In relayer v2 we should add a separate `ErrorResponse` type.
- Clean up Substrate/Evm command types so they can be checked at compile time and dont throw any runtime error.

- Closes 

-----
### Code Checklist 

- [x] Tested
- [x] Documented

### DIFF
--- a/crates/relayer-handlers/src/lib.rs
+++ b/crates/relayer-handlers/src/lib.rs
@@ -34,8 +34,8 @@ use webb_proposals::TypedChainId;
 
 use webb_relayer_context::RelayerContext;
 use webb_relayer_handler_utils::{
-    Command, CommandResponse, CommandStream, CommandType, EvmCommand,
-    IpInformationResponse, SubstrateCommand,
+    Command, CommandResponse, CommandStream, EvmCommandType,
+    IpInformationResponse, SubstrateCommandType,
 };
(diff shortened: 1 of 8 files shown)


crates/relayer-handlers/src/lib.rs
```diff
@@ -189,7 +189,13 @@ pub async fn handle_evm(
     stream: CommandStream,
 ) {
     if let CommandType::VAnchor(_) = cmd {
-        handle_vanchor_relay_tx(ctx, cmd, stream).await
+        match handle_vanchor_relay_tx(ctx, cmd, stream.clone()).await {
+            Ok(o) => o,
+            Err(e) => {
+                tracing::error!("{:?}", e);
+                let _ = stream.send(e);
```
Nutomic, line 196: Instead of having error loggin and stream.send all over the code, we can do all of it here which makes the code much easier to read. If other methods handle errors in the same way, this can be moved even further up the stack.

shekohex (approved):
LGTM!

PR #335, opened 2023-01-31
WIP: Get the Eth2 Substrate hooked in to the relayer

tbraun96:


.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="VcsDirectoryMappings">
+    <mapping directory="" vcs="Git" />
+  </component>
+</project>
```
Nutomic, line 6: You should remove .idea folder and add it to .gitignore.

Nutomic: Nevermind I did that in my PR.

services/light-client-relayer/tests/test_client.rs
```diff
@@ -0,0 +1,5 @@
+#[cfg(test)]
```
drewstone, line 1: Is this file needed or is there a plan to use it somehow?

drewstone:
What's the status of this @tbraun96?

drewstone (changes requested):

tbraun96:
Any ideas why the integration test is failing for `x86_64` linux but not `aarch64` linux?

shekohex:
> Any ideas why the integration test is failing for `x86_64` linux but not `aarch64` linux?

Simply, since it does not run on `aarch64`, we skip it.

tbraun96:
Looks like we're getting new errors now?

shekohex:
I changed this to target the `develop` branch, @tbraun96 can you resolve these conflicts? Looks like it is just in the Cargo.lock (delete & regenerate may fix it)

PR #340, opened 2023-02-07
Fix data overwrite and refresh nonce issue in event watcher

### LINKED ISSUE #339: [BUG]  Data Overwrite Issue with Event Watchers
## Overview

Currently implementation of  event watchers uses `ResourceId` key as `HistoryStoreKey` to access and store data  which will be same for all events watcher for given TargetSytem. This causes data overwrite Issue  where one event watcher overwrites `BlockNumber` saved by other resulting in misbehavior. 

`HistoryStoreKey` is an enum of keys used to read/write MemoryStore in relayer and we can use ResourceId as key
```rust
/// HistoryStoreKey contains the keys used to store the history of events.
#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
pub enum HistoryStoreKey {
    /// Block Queue Key
    Block { chain_id: u32 },
    /// ResourceId Key
    ResourceId {
        resource_id: webb_proposals::ResourceId,
    },
}
```

Current Architecture
- Every event watcher instance fetches block details and processes it individually
- Every event watcher uses  same `HistoryStoreKey` to save processed Block Number from MemoryStore 

## Solution 1
We make sure `HistoryStoreKey` is unique for each event watcher(depoit,leaves,encrypted, signatureBridge) 

```rust
pub enum HistoryStoreKey {
  resource_id: webb_proposals::ResourceId,
  identifier: <unique_identifer>
}
```

## Solution 2

We refactor event_watcher. We run one event watcher for each resourceID which will consume multiple event_handlers.

```rust

pub struct SubstrateEventWatcher<H:SubstrateEventHandler>{
    event_handlers: Vec<H>
}

impl <H> SubstrateEventWatcher<H> {
    const TAG: &'static str = "Substrate bridge pallet Watcher";
     /// Returns a task that should be running in the background
    /// that will watch events
    async fn run(
        &self,
        _node_name: String,
        chain_id: u32,
        client: Arc<OnlineClient<subxt::SubstrateConfig>>,
        store: Arc<SledStore>,
        metrics: Arc<Mutex<metric::Metrics>>,
    ) -> webb_relayer_utils::Result<()> {
    loop {

                let history_store_key =
                    ResourceId::new(src_target_system, src_typed_chain_id);
                // get block number
                let block = store
                    .get_last_block_number(history_store_key, 1u64)
                // get events for block
                let events = client
                    .events()
                    .at(Some(block))

                for event_handler in event_handlers {
                    let result = event_handler
                    .handle_event(
                        store.clone(),
                        client.clone(),
                        (events, block_number),
                        metrics_clone.clone(),
                    )
                    .await;
                    match result {
                        Ok(_) => {
                            
                            tracing::trace!(
                                "event handled successfully. at #{}",
                                current_block_number
                            );
                        }
                        Err(e) => {
                            tracing::error!(
                                "Error while handling event: {}",
                                e
                            );
                            
                            return Err(backoff::Error::transient(e));
                      }

                }

                let current_block_number: u64 =  block_number.into();

                store.set_last_block_number(
                    history_store_key,
                    current_block_number,
                )?;
            }
       }
   }


/// Represents a Substrate event watcher.
#[async_trait::async_trait]
pub trait SubstrateEventHandler {
    /// A helper unique tag to help identify the event watcher in the tracing logs.
    const TAG: &'static str;
    /// The Config of this Runtime [`subxt::PolkadotConfig`, `subxt::SubstrateConfig`]
    type RuntimeConfig: subxt::Config + Send + Sync + 'static;
    /// The Runtime Client that can be used to perform API calls.
    type Client: OnlineClientT<Self::RuntimeConfig> + Send + Sync;
    /// All types of events that are supported by this Runtime.
    /// Usually it will be [`my_runtime::api::Event`] which is an enum of all events.
    type Event: scale::Decode + Send + Sync + 'static;
    /// The kind of event that this watcher is watching.
    type FilteredEvent: subxt::events::StaticEvent + Send + Sync + 'static;
    /// The Storage backend, used by the event watcher to store its state.
    type Store: HistoryStore;

    /// A method to be called with the event information,
    /// it is up to the handler to decide what to do with the event.
    /// If this method returned an error, the handler will be considered as failed and will
    /// be retried again, depends on the retry strategy.
    async fn handle_event(
        &self,
        store: Arc<Self::Store>,
        client: Arc<Self::Client>,
        (event, block_number): (Self::FilteredEvent, BlockNumberOf<Self>),
        metrics: Arc<Mutex<metric::Metrics>>,
    ) -> webb_relayer_utils::Result<()>;

}



```

salman01zp:
## Summary of changes
- [x] Fix data overwrite Issue with event watcher
- [x] Add refresh nonce checks, similar to solidity
- [x] Update webb-rs crate

- Closes #339 

-----
### Code Checklist 

- [x] Tested
- [ ] Documented

### DIFF
--- a/crates/event-watcher-traits/src/substrate/event_watcher.rs
+++ b/crates/event-watcher-traits/src/substrate/event_watcher.rs
@@ -22,6 +22,8 @@ use super::*;
 pub trait SubstrateEventWatcher {
     /// A helper unique tag to help identify the event watcher in the tracing logs.
     const TAG: &'static str;
+    /// Pallet name  used to fetch pallet id for given target system
+    const PALLET_NAME: &'static str;
     /// The Config of this Runtime [`subxt::PolkadotConfig`, `subxt::SubstrateConfig`]
     type RuntimeConfig: subxt::Config + Send + Sync + 'static;
     /// The Runtime Client that can be used to perform API calls.
@@ -69,6 +71,15 @@ pub trait SubstrateEventWatcher {
             let mut instant = std::time::Instant::now();
             let step = 1u64;
             let rpc = client.rpc();
+            // get pallet index
+            let pallet_index = {
+                let metadata = client.metadata();
+                let pallet = metadata
+                    .pallet(Self::PALLET_NAME)
+                    .map_err(Into::into)
+                    .map_err(backoff::Error::permanent)?;
+                pallet.index()
+            };
             loop {
                 // now we start polling for new events.
                 // get the latest seen block number.
@@ -76,7 +87,7 @@ pub trait SubstrateEventWatcher {
                 // create history store key
                 let src_typed_chain_id = TypedChainId::Substrate(chain_id);
                 let target = SubstrateTargetSystem::builder()
-                    .pallet_index(chain_id as u8)
+                    .pallet_index(pallet_index)
                     .tree_id(chain_id)
                     .build();
                 let src_target_system = TargetSystem::Substrate(target);
@@ -115,6 +126,11 @@ pub trait SubstrateEventWatcher {
                 let dest_block = cmp::min(block + step, current_block_number);
                 // check if we are now on the latest block.
                 let should_cooldown = dest_block == current_block_number;
+                tracing::debug!(
+                    %dest_block,
+                    %current_block_number,
+                    %block
+                );
                 tracing::trace!("Reading from #{} to #{}", block, dest_block);
                 // Only handle events from found blocks if they are new
                 if dest_block != block {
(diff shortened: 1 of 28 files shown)


salman01zp:
will fix CI tests

crates/event-watcher-traits/src/substrate/event_watcher.rs
```diff
@@ -115,14 +126,19 @@ pub trait SubstrateEventWatcher {
                 let dest_block = cmp::min(block + step, current_block_number);
                 // check if we are now on the latest block.
                 let should_cooldown = dest_block == current_block_number;
+                tracing::debug!(
+                    %dest_block,
+                    %current_block_number,
+                    %block
+                );
```
shekohex, line 133:
```suggestion
                tracing::trace!(
                    %dest_block,
                    %current_block_number,
                    %block
                );
```

shekohex (approved):
LGTM!

PR #343, opened 2023-02-10
Some adjustments for refund code

Nutomic:
## Summary of changes
- `FeeInfo.max_refund` is now in native token
- Relayer checks that it has enough balance for refund
- Actually make refund work

- Closes 

-----
### Code Checklist 

- [ ] Tested
- [ ] Documented


crates/tx-relay/src/evm/vanchor.rs
```diff
@@ -163,8 +180,13 @@ pub async fn handle_vanchor_relay_tx<'a>(
 
     // check the fee
     // TODO: This adjustment could potentially be exploited
-    let adjusted_fee = fee_info.estimated_fee / 100 * 98;
-    if cmd.ext_data.fee < adjusted_fee {
+    let adjusted_fee = fee_info.estimated_fee / 100 * 96;
```
Nutomic, line 183: Had to increase the adjustment amount. This is because gas estimation in Typescript test is done without refund. I tried adding the refund for more accurate gas estimate, but that results in error "Incorrect refund amount received by the contract" when `dummyOutput` is constructed.

shekohex: Ah Makes sense, sounds good for now.

PR #346, opened 2023-02-13
Felix/anvil tests

Nutomic:
## Summary of changes
- Added anvil as dev dependency. This is pretty messy, the only working tag I found in their repo requires a dev version of ethers.

- Closes 

-----
### Code Checklist 

- [ ] Tested
- [ ] Documented


Nutomic:
So I got anvil running but dont have any success deploying a token. Calling the deploy function always results in error "Out of gas: required gas exceeds allowance: 10000000000000000000". The amount mentioned is always identical to the `gas_limit` configured for anvil, even if I set it to 100 eth. I dont think its really related to allowance as that can only be set after the token is deployed. Maybe one of you has an idea. You can run the test with `cargo test -p webb-relayer-tx-relay`.

PR #347, opened 2023-02-13
sync blocks from config

### LINKED ISSUE #345: [TASK] Add a Start Block for syncing in Substrate configration
### Overview

Right now relayers do start syncing on the substrate, they sync from block zero, which could take more time than needed. We need to add a configuration option to allow for custom starting block number, similar to what we are doing in the evm contracts `deployed_at` value.

### Example

```toml
[substrate.tangle]
name = "tangle"
http-endpoint = "http://localhost:9933"
ws-endpoint = "ws://localhost:9944"
chain-id = 1080
enabled = true
runtime = "DKG"
suri = "//Bob"
start-block = 1000 # if not provided, we start from the latest block.

[[substrate.tangle.pallets]]
pallet = "DKG"
events-watcher = { enabled = true, polling-interval = 3000, print-progress-interval = 30000 }

[[substrate.tangle.pallets]]
pallet = "DKGProposalHandler"
events-watcher = { enabled = true, polling-interval = 3000, print-progress-interval = 30000 }
```

salman01zp:
## Summary of changes
- Add config to set custom block number for a relayer to start syncing
```rust
     pub struct EventsWatcherConfig {
          /// if it is enabled for this chain or not.
          pub enabled: bool,
          .
          /// Sync blocks from
          #[serde(rename(serialize = "syncBlocksFrom"))]
          pub sync_blocks_from: Option<u64>,
     }
```
  ```rust
    let sync_blocks_from: u64 = event_watcher_config
          .sync_blocks_from
          .unwrap_or(current_block_number);
    // get latest saved block number
    let block = store
          .get_last_block_number(history_store_key, sync_blocks_from)
          .map_err(Into::into)
          .map_err(backoff::Error::transient)?;
  ```
- Use `print-progress-interval` from config

- Closes #345 

-----
### Code Checklist 

- [ ] Tested
- [ ] Documented

### DIFF
--- a/crates/event-watcher-traits/src/substrate/event_watcher.rs
+++ b/crates/event-watcher-traits/src/substrate/event_watcher.rs
@@ -121,7 +121,7 @@ pub trait SubstrateEventWatcher {
                     current_block_number
                 );
                 let sync_blocks_from: u64 = event_watcher_config
-                    .sync_blocks_form
+                    .sync_blocks_from
                     .unwrap_or(current_block_number);
                 // get latest saved block number
                 let block = store
--- a/crates/event-watcher-traits/src/substrate/event_watcher.rs
+++ b/crates/event-watcher-traits/src/substrate/event_watcher.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use tokio::sync::Mutex;
+use webb_relayer_config::event_watcher::EventsWatcherConfig;
 use webb_relayer_utils::metric;
 
 use super::*;
@@ -59,11 +60,11 @@ pub trait SubstrateEventWatcher {
     )]
     async fn run(
         &self,
-        _node_name: String,
         chain_id: u32,
         client: Arc<Self::Client>,
         store: Arc<Self::Store>,
         metrics: Arc<Mutex<metric::Metrics>>,
+        event_watcher_config: EventsWatcherConfig,
     ) -> webb_relayer_utils::Result<()> {
         let backoff = backoff::backoff::Constant::new(Duration::from_secs(1));
         let metrics_clone = metrics.clone();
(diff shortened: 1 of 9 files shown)


crates/event-watcher-traits/src/substrate/event_watcher.rs
```diff
@@ -222,7 +228,12 @@ pub trait SubstrateEventWatcher {
                     tokio::time::sleep(duration).await;
                 }
                 // only print the progress if 7 seconds (by default) is passed.
```
shekohex, line 230:
```suggestion
```

shekohex (approved):
LGTM!

shekohex:
@salman01zp Could you also please update the `README` file with the new config options we added?

crates/relayer-config/src/event_watcher.rs
```diff
@@ -18,6 +18,12 @@ pub struct EventsWatcherConfig {
...
     /// print sync progress frequency in milliseconds
     /// if it is zero, means no progress will be printed.
-    #[serde(skip_serializing, default = "print_progress_interval_default")]
+    #[serde(
+        rename(serialize = "printProgressInterval"),
+        default = "print_progress_interval_default"
+    )]
     pub print_progress_interval: u64,
+    /// Sync blocks from
+    #[serde(rename(serialize = "syncBlocksFrom"))]
```
Nutomic, line 27: Why are these two in a different case? (camelCase instead of kebab-case like the rest)

Nutomic (approved):

crates/relayer-config/src/event_watcher.rs
shekohex: these are for the response sent to the users when querying to `/api/v1/info` endpoint, these are most camelCase for the dApp.

salman01zp: If I am passing values from typescript which is in camelCase, while other options have default values.

Nutomic: Then shouldnt this entire struct use `#[serde(rename_all = "camelCase")]`? Though that would be a breaking change.

shekohex: Yeah `#[serde(rename_all = "camelCase")]` would be a breaking change, however we only need it in `camelCase` when we serialize only.

PR #349, opened 2023-02-15
Use workspace inheritance (fixes #348)

### LINKED ISSUE #348: [TASK] Workspace inheritance
Rust since 1.64 supports "workspace inheritance", which means that dependencies and metadata are only specified once in the top-level Cargo.toml file. This avoids duplication and makes changes much easier, for example if a specific dependency version needs to be changed.

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#inheriting-a-dependency-from-a-workspace

Nutomic:
## Summary of changes
- Use workspace inheritance to specify dependencies and crate metadata only one. I only extracted deps which are used by two or more crates. Regarding crate versions I wasnt sure if its intentional that they differ. If all crates should have the same version then this should also be specified in the root `Cargo.toml`.

By the way, in my opinion this project is split into too many crates. I dont think there is any benefit to having crates with only a single file.

- Closes #348

-----
### Code Checklist 

- [ ] Tested
- [ ] Documented

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6326,7 +6326,6 @@ dependencies = [
  "dotenv",
  "ethereum-types 0.14.1",
  "futures",
- "native-tls",
  "paw",
  "serde_json",
  "tempfile",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,10 +1,10 @@
-[profile.release]
-strip = "symbols"
-lto = "thin"
-
-[profile.dev]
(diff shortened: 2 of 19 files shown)


drewstone (approved):
I do really like this feature! Maybe we can introduce this in all our rust repos?

Nutomic:
@drewstone Alright I can take care of that.

shekohex (approved):

PR #355, opened 2023-03-02
Bridge Registry pallet integration

### LINKED ISSUE #197: [SPEC] Remove state of linked anchors from relayer configs
(this shouldn't be taken on right now)

# Overview
Relayers don't need to store the linked anchor config, instead we can get this information from the edge list of a deployed anchor. The main challenge is to handle dynamic additions to the bridge. The relayer would need to create a new event watcher on the fly if it hasn't set up a watcher for a new addition to an anchor's edge list.

The new anchor addition flow would also required hooking up the `AnchorUpdateProposal` to the `force_submit_unsigned_proposal` so that we can get it signed for inclusion into all existing anchors. Finally, we must finish the loop by updating the new anchor with the state of all existing anchors as well.

# Checklist
- [ ] Spec a dynamic addition of a new anchor onto an existing bridge
- [ ] Remove configuration and use on-chain
- [ ] Define flow for adding new anchors to an existing bridge
- [ ] Spawn new event watchers when new anchors are added to an existing bridge (and in general)
- [ ] more to spec...

### LINKED ISSUE #358: [TASK] Bridge Registry pallet integration v1
### Overview

The purpose of this document is to outline the features and intended functionality that will ship bridge registry pallet support on the relayer.

- Bridge Registry pallet will make it easier for the relayer operators to configure bridges on the fly.
- This will eliminate the need for `linked-anchor` configuration files.

### Features

- On the Startup, Query the `Bridges` from the `BridgeRegistry` pallet.
- Create a graph of linked bridges, by parsing the `ResourceId`s and known Anchors.
- Replace the current logic for Linked Anchor configurations with the one fetched from the pallet.

### Config Files changes

- [x] make the `linked-anchors` field optional in the config. (_currently, we set the default to be empty-list, we need it to also be optional, if `None` we will build it_)

### Checklist

- [x]  Explore the Bridge Registry pallet
- [x]  Run a node that has the bridge registry pallet configured.
- [x]  Send unsigned `AnchorUpdate` proposals to the DKG to be signed.
- [x]  Read the Bridges from the pallet.
- [x]  Create a simple script/program in Rust using subxt and webb sdk to send proposals and read the bridges from the pallet.
- [x]  Integrated the above logic inside the relayer.
- [x]  Remove the need for the Linked Anchors inside the config files
- [x]  Update integration tests and docs with the new changes.

Nutomic:
You can execute this with `cargo test -p webb-proposal-signing-backends -- --nocapture`. `./scripts/run-standalone.sh` from dkg-substrate repo must be running.

## Summary of changes
- Add a test case which sends unsiged `AnchorUpdate` to DKG to be signed (converted from [DKG submitProposals.ts test](https://github.com/webb-tools/dkg-substrate/blob/master/dkg-test-suite/scripts/submitProposals.ts)
- Add bridge registry backend for calling DKG bridge registry pallet methods
- Use this pallet to retrieve linked anchors from DKG if none are configured in config file
- Didnt change anything in config files at this time to avoid conflicts

- Closes #197
- Closes #358 

-----
### Code Checklist 

- [x] Tested
- [x] Documented


crates/proposal-signing-backends/tests/submit_anchor_update_proposal.rs
```diff
@@ -0,0 +1,164 @@
...
+
+    println!("register resource 1");
+    let register_resource1 =
+        tx_api.set_resource(DkgResourceId(resource_id.into_bytes()), vec![]);
+    let mut progress = api
+        .tx()
+        .sign_and_submit_then_watch_default(&register_resource1, &sudo_account)
+        .await
+        .unwrap();
+    watch_events(&mut progress).await;
```
Nutomic, line 71: This is failing with `Runtime error: Other([2, 198, 228, 152, 23, 0, 0, 0])`. If I register the resources manually via Polkadot js admin ui and comment out this code, the test passes.

shekohex:
fyi, @Nutomic I did change the base branch to be `develop`.

services/webb-relayer/src/service.rs
```diff
@@ -660,7 +662,11 @@ async fn start_evm_vanchor_events_watcher(
         .await?;
         match proposal_signing_backend {
             ProposalSigningBackendSelector::Dkg(backend) => {
-                let deposit_handler = VAnchorDepositHandler::new(backend);
+                let bridge_registry = DkgBridgeRegistryBackend::new(
+                    OnlineClient::<PolkadotConfig>::new().await?,
+                );
```
Nutomic, line 667: I didnt create a separate init method for bridge registry because it would be more trouble than its worth for this little code. Especially with another dkg/mocked selector the match would get very complicated. This does the job fine (unless we want to pass in more parameters).

Cargo.toml
```diff
@@ -41,7 +39,7 @@ tokio = { version = "^1", features = ["full"] }
 config = { version = "0.13", default-features = false, features = ["toml", "json"] }
 serde_json = { version = "^1", default-features = false }
 paw = { version = "^1.0" }
-webb = { version = "0.5.17", default-features = false }
+webb = { git = "https://github.com/webb-tools/webb-rs.git", rev = "2bbe7b6a4f9", default-features = false }
```
shekohex, line 42: We will publish the new version of webb-rs with the new changes, so we will hold this PR until we update the runtime of protocol-substrate there too.

cc @salman01zp are there a PR to do that already?

crates/bridge-registry-backends/src/dkg.rs
```diff
@@ -0,0 +1,63 @@
...
+            .client
+            .storage()
+            .at(None)
+            .await?
+            .fetch(&next_bridge_index)
+            .await?
+            .unwrap_or(1))
+    }
+
+    async fn resource_to_bridge_id(
```
shekohex, line 31: Let's be more clear about naming methods, this uses an inner method called `resource_to_bridge_index` so lets use that.

```suggestion
    async fn resource_to_bridge_index(
```

crates/bridge-registry-backends/src/dkg.rs
```diff
@@ -0,0 +1,63 @@
...
+        &self,
+        resource_id: &ResourceId,
+    ) -> webb_relayer_utils::Result<u32> {
```
shekohex, line 34: This method should return an `Option<u32>` and if it is `None` that means this bridge does not exist. Using this method right now, as you can see it will return index of `1` which means bridge 1, this resource could not be in that bridge at all, resulting in a bug.

crates/bridge-registry-backends/tests/submit_anchor_update_proposal.rs
```diff
@@ -0,0 +1,118 @@
...
+use webb_bridge_registry_backends::BridgeRegistryBackend;
+use webb_proposals::evm::AnchorUpdateProposal;
+use webb_proposals::ResourceId;
+use webb_proposals::{FunctionSignature, Nonce, ProposalHeader};
+
+/// Rust reimplementation of DKG submit proposal test
+///
+/// <https://github.com/webb-tools/dkg-substrate/blob/master/dkg-test-suite/scripts/submitProposals.ts>
+#[tokio::test]
+async fn submit_anchor_update_proposal() {
```
Nutomic, line 20: Not sure if we can execute this test in CI as it requires a running relayer. At least its already checking that the code here compiles.

crates/bridge-registry-backends/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+pub trait BridgeRegistryBackend {
+    /// Returns the next available bridge index. In other words, it returns the index of the
+    /// last bridge + 1.
+    async fn next_bridge_index(&self) -> webb_relayer_utils::Result<u32>;
+
+    /// Retrieves the bridge ID which a given resource is attached to.
+    ///
+    /// The passed resource must belong to a bridge that is registered with DKG, otherwise an error
+    /// is returned.
+    async fn resource_to_bridge_id(
```
shekohex, line 49:
```suggestion
    async fn resource_to_bridge_index(
```

crates/bridge-registry-backends/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+pub mod mocked;
+
+/// Wrapper for calling DKG bridge registry pallet methods.
+#[async_trait::async_trait]
+pub trait BridgeRegistryBackend {
+    /// Returns the next available bridge index. In other words, it returns the index of the
+    /// last bridge + 1.
+    async fn next_bridge_index(&self) -> webb_relayer_utils::Result<u32>;
+
+    /// Retrieves the bridge ID which a given resource is attached to.
```
shekohex, line 45:
```suggestion
    /// Retrieves the bridge index which a given resource is attached to.
```

crates/bridge-registry-backends/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+            None => {
+                let next_bridge_index =
+                    self.resource_to_bridge_id(linked_anchor).await?;
+                let bridges = self.bridges(next_bridge_index).await?.unwrap();
+                Ok(bridges
+                    .resource_ids
+                    .0
+                    .into_iter()
+                    .filter(|r| r.0 != linked_anchor.0)
+                    .map(|r| {
```
shekohex, line 84: Use [`filter_map`](https://doc.rust-lang.org/stable/core/iter/trait.Iterator.html#method.filter_map) instead

shekohex (changes requested):
Looks good so far, a few comments about the terminology

crates/bridge-registry-backends/tests/submit_anchor_update_proposal.rs
shekohex: just add `#[ignore = "requires ...."]` on that.

crates/bridge-registry-backends/src/lib.rs
Nutomic: That might be a good idea in some cases but here it requires more code because of explicit if else. Current implementation is simpler.

shekohex: Sounds good!

Nutomic:
Updated for review comments, but now CI is throwing weird internal compiler errors. It passed previously and the changes were very minor. So Im guessing the nightly is broken.

salman01zp:
Yes latest nightly is failing I am using [nightly-2023-03-12](https://github.com/webb-tools/relayer/pull/371/commits/a55d1bc605902d2e2f72ffb92b8bdcd6e9171a3d#diff-2b1bde2cf3a858b7bf7424cb8bcbf01f35b94dc80b925d9432cbab3319ca9b4eR2) for now

shekohex:
> Yes latest nightly is failing I am using [nightly-2023-03-12](https://github.com/webb-tools/relayer/pull/371/commits/a55d1bc605902d2e2f72ffb92b8bdcd6e9171a3d#diff-2b1bde2cf3a858b7bf7424cb8bcbf01f35b94dc80b925d9432cbab3319ca9b4eR2) for now

@Nutomic can you start using this and push that so that CI works?

Cargo.toml
shekohex: @Nutomic could you update this and use the new version of webb?

salman01zp:
we also need to update validation checks [here](https://github.com/webb-tools/relayer/blob/main/services/webb-relayer/src/service.rs#L1202-#L1215) . 
If the relayer is configured with a governance feature enabled, it will validate if linked anchors are provided before starting the governance relaying service

shekohex (approved):

PR #365, opened 2023-03-09
relayer transfer test

### LINKED ISSUE #353: [TASK] Add Support for the `transfer` flow using the Relayer
## Overview

The transfer flow is that we can do a transfer between two parties without the assets moving out of the pool, this so useful as it would work without going through the `Deposit` -> `Withdraw` then another `Deposit` tx from the other peer.

If we have two parties, Alice and Bob, Alice can transfer some already wrapped assets from the pool to Bob, without doing a withdraw to Bob and then Bob have to do another deposit call. this already implemented to happen in one call.

To maximize the privacy, Alice should use a relayer to do this tx instead of her, to cut any links between her and Bob.

## Task Checklist

- [ ] Implement the transfer flow in the relayer with the already supported withdraw flow. (we can rename it to `transact` in the future).
- [ ] Add test for the transfer flow, similar to the withdraw flow.

Examples in tests:
1. https://github.com/webb-tools/protocol-solidity/blob/118ea3acf403e4693dbc4b8457eb7435885deb02/packages/contracts/test/vanchor/vanchor.test.ts#L491
2. https://github.com/webb-tools/protocol-solidity/blob/118ea3acf403e4693dbc4b8457eb7435885deb02/packages/contracts/test/vanchor/vanchor.test.ts#L664

salman01zp:
## Summary of changes
- Add transfer test using relayer

- Closes #353 

-----
### Code Checklist 

- [ ] Tested
- [ ] Documented


shekohex (approved):
LGTM!

PR #372, opened 2023-03-20
Add etherscan api configuration

### LINKED ISSUE #370: [TASK] Add `evm-etherscan` configuration section to the Relayer Config
### Overview

Currently, the relayer will panic during the startup if there is no `ETHERSCAN_API` env present, which is not something we want at all. However, we should add a configuration section that is required if the user did enable the `private-tx-relay` feature.

https://github.com/webb-tools/relayer/blob/1e7bb7f51cf666eedd4df18a922694cdf8f88ed4/crates/relayer-context/src/lib.rs#L71

Clippy does a great job here showing all the panic paths:
image

### Proposed Change

In the `main.toml` config file, the user should define the following section:
```toml
# For Etherum mainnet
[evm-etherscan.mainnet]
chain-id = 1
api-key = "$ETHERSCAN_MAINNET_API_KEY"
# For Polygon
[evm-etherscan.polygon]
chain-id = 137
api-key = "$POLYGONSCAN_MAINNET_API_KEY"
```

That would translate in our config struct to something like the following:
```rust
struct WebbRelayerConfig {
	// ...
	evm_etherscan: HashMap<String, EtherscanApiConfig>,
	// ...
}

struct EtherscanApiConfig {
	chain_id: u64,
	/// A wrapper type around the `String` to allow reading it from the env.
	api_key: EtherscanApiKey,
}
```

## Checklist

- [ ] Add the `evm_etherscan` to the `WebbRelayerConfig`
- [ ] Create the `EtherscanApiConfig` struct and the `EtherscanApiKey` type.
- [ ] Add the default values to be empty map of the `evm_etherscan`.
- [ ] Update the `RelayerContext` to have a mapping between the `ChainId` <=> `EtherscanApiClient`
- [ ] Update the use of the old etherscan client to start using the new mapping the context
- [ ] Return an Error if that chain does not have a corresponding etherscan client.
- [ ] Update the `README.md` and old configuration files with the new changes.

salman01zp:
## Summary of changes
- [x] Add configuration for `EtherscanApi`
- [x] Update tests to configure `EtherscanApi` for chain

- Closes #370 

-----
### Code Checklist 

- [x] Tested
- [x] Documented


dutterbutter:
@salman01zp friendly reminder any changes to the config should be reflected in the docs both in the repo and the docs site.

shekohex:
Could you also update the docs? the https://github.com/webb-tools/relayer/blob/053062d4c2af5af970fbffae87c352f7aa19e56b/config/README.md needs to be updated with the new changes.

drewstone (approved):
LG!

PR #375, opened 2023-03-21
Fees for substrate

### LINKED ISSUE #364: [TASK] Calculate estimated fee for substrate based transactions
## Overview
Calculate the estimated fee for relaying substrate based transactions/extrinsic. Fees on substrate based chains are calculated based on three parameters
- Base Fee
- Length Fee
- Network congestion

You can read more about fee details [here](https://wiki.polkadot.network/docs/learn-transaction-fees) 

Here is an open issue on subxt [Return fee details for an Extrinsic #527](https://github.com/paritytech/subxt/issues/527)

-----
### Task Checklist
Similar to EVM Fees Enpoint, we will need one for Substrate.

- [ ] Add an API endpoint that takes in the (TypedChainId, VAnchorTreeId, estimatedXtFees) and returns the relayer estimated fees.
- [ ] Calculate the estimated fee for relaying transaction (see below for the calculation)
- [ ] Add validation check for fee submitted by a user.
- [ ] Write tests to ensure that the above functionality works correctly

### Estimate the Extrinsic fees

There is an RPC method that you can call with the signed extrinsic call and it will return the fees details, this RPC is `payment_queryFeeDetails`. Here is an example:

* Request Payload:
```json
{"id":1,"jsonrpc":"2.0","method":"payment_queryInfo","params":["0xad0184008eaf04151687736326c9fea17e25fc5287613693c912909cb226aa4794f26a48010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010124010000000004306d000000"]}
```
* Response:

```json
{"jsonrpc":"2.0","result":{"weight":1307595,"class":"normal","partialFee":"1130000000100090506"},"id":1}
```

On the client side, the user will use this method to estimate the fees so they can send to the relayer. 

You will notice the following information in the result:

```js
partialFee = "1130000000100090506";
```

The final fee is like the following:
```js
// Calculates the fees in how much tokens
fee = parseInt(partialFee) / (10 ** tokenDecimals)
```

Example:
```js
tokenDecimals = 18; // for tTNT (Tangle Standalone Test Network)
partialFee = "1130000000100090506";
fee = parseInt(partialFee) / (10 ** tokenDecimals) // => 1.1300 tTNT

// The calculated fees above, is the base transaction fees that the relayer will pay for sending this transaction.
```

If you are using polkadot API, you can see the docs [here](https://polkadot.js.org/docs/api/cookbook/tx#how-do-i-estimate-the-transaction-fees) about estimating the transaction fees.

### Wrapped Token to the Native token Conversion

When working on the VAnchors, we are working on wrapped tokens, each VAnchor as an associated `assetId` that can be later used to get that asset information, for example say we have `assetId = 1`, by calling `assetRegistry.assets` we will get the following:
```
{
  name: webbtTNT
  assetType: PoolShare
  existentialDeposit: 100,000,000,000,000,000
  locked: false
}
```

You will notice the `name` is `webbtTNT` and it is a wrapped token over the native `tTNT` token, so it should be `1:1` conversion in this case.

> Please Notice that since currently there is no way for us to fetch prices of Substrate tokens, we will (for now) hardcode the price of the `1 tTNT = $0.10`. That also could change in the future.

### LINKED ISSUE #503: [ISSUE] Getting InvalidTransaction proof error on vanchor withdraw
## Overview

The Vanchor test generates invalid transaction proof when using the tangle node.

```js
const isValidProof = verify_js_proof(
    data.proof,
    data.publicInputs,
    u8aToHex(vk).replace('0x', ''),
    'Bn254'
  );
console.log('Is withdrawal proof valid : ', isValidProof);
  ```
 All inputs seem okay, following where inputs are pulled from on-chain chainId, leaves, and root. 
 
 After cross-checking found out that we were using the wrong default leaf value [here](https://github.com/webb-tools/tangle/blob/main/standalone/runtime/src/protocol_substrate_config.rs#L48) due to which invalid root hash in generated causing proof to be invalid.
 
## Task Checklist
- [x] Update verifying keys on the tangle
- [x] Update default leaf value

### PR DESCRIPTION
**Pull request by Nutomic**
## Summary of changes
- Add fee and refund logic for substrate
- Update test instructions to include tangle setup
- Update fee endpoints
- Update test case `should withdraw using private transaction` to use fee and refund

## Note
We have updated the fee endpoint to support requests for  substrate and evm based chains, following are the new points
- EVM:
```
  /api/v1/fee_info/evm/:chain_id/:vanchor/:gas_amount
```
-  Substrate
```
 /api/v1/fee_info/substrate/:chain_id/:partialFee
```

- Closes #364 
- Closes #503 
-----
### Code Checklist 

- [ ] Tested
- [ ] Documented

### CODE REVIEW

File: crates/tx-relay/src/evm/fees.rs [hunk 1]
```diff
@@ -153,11 +156,15 @@ async fn generate_fee_info(
             .into();
 
     // Calculate the maximum refund amount per relay transaction in `nativeToken`.
+    // Ensuring that refund <= relayer balance
+    let relayer_balance =
+        relayer_balance(chain_id.chain_id(), ctx).await.unwrap();
     let max_refund =
         parse_units(MAX_REFUND_USD / native_token_price, native_token.1)?
             .into();
+    let max_refund = min(relayer_balance, max_refund);
```
Nutomic, line 165: Noticed that this check was previously missing in evm fee code, so a client might request refund that is higher than relayer balance.

crates/tx-relay/src/substrate/fees.rs
```diff
@@ -0,0 +1,65 @@
...
+    let refund_exchange_rate = matic_to_wei(1.);
+    // TODO: should ensure that refund <= relayer balance
+    let max_refund = matic_to_wei(MAX_REFUND_USD / TOKEN_PRICE_USD);
+    SubstrateFeeInfo {
+        estimated_fee,
+        refund_exchange_rate,
+        max_refund,
+        timestamp: Utc::now(),
+    }
+}
```
Nutomic, line 41: Very simple logic because its using hardcoded values. For this reason I also havent implemented any caching.

crates/tx-relay/src/substrate/vanchor.rs
```diff
@@ -74,17 +82,58 @@ pub async fn handle_substrate_vanchor_relay_tx<'a>(
...
 
-    let signer = PairSigner::new(pair);
+    let signer = PairSigner::new(pair.clone());
 
+    // TODO: add refund
     let transact_tx = RuntimeApi::tx().v_anchor_bn254().transact(
         cmd.id,
         proof_elements,
         ext_data_elements,
     );
```
Nutomic, line 92: Not sure how to send the refund in substrate.

crates/tx-relay/src/substrate/vanchor.rs
```diff
@@ -74,17 +82,58 @@ pub async fn handle_substrate_vanchor_relay_tx<'a>(
...
+        .unwrap();
+    dbg!(&payment_info);
+    let fee_info =
+        get_substrate_fee_info(requested_chain, payment_info.partial_fee, &ctx)
+            .await;
+
+    // TODO: check refund amount <= relayer wallet balance
+    let mut params = RpcParams::new();
+    params.push(hex::encode(pair.public().deref())).unwrap();
+    //let balance = client.storage().at(None).await.unwrap().
```
Nutomic, line 114: I cant manage to get the relayer wallet balance here.

crates/tx-relay/src/substrate/vanchor.rs
shekohex: it is in the `ext_data_elements` already set by the user, and the pallet automatically handles the native token transfer for you.

crates/tx-relay/src/substrate/vanchor.rs
shekohex: I guess we resolved that in our conversation on Slack, no?

Nutomic: No this is still not working, its returning None. I pushed my latest code just now.

Nutomic:
Currently waiting for this PR to make webb.js dependencies work with Node 18.

https://github.com/webb-tools/webb.js/pull/311

drewstone:
This can be updated now afaik

shekohex:
> This can be updated now afaik

Updated and fixed the merge conflicts too.

Nutomic:
I pushed the changes made with @shekohex to fix fee calculation. Now its still failing with the error `Error: the string "Error while watching Tx: Runtime error: Module(ModuleError { pallet: \"VAnchorBn254\", error: \"InvalidExtData\", description: [], error_data: ModuleErrorData { pallet_index: 41, error: [6, 0, 0, 0] } })" was thrown, throw an Error :)`. Most likely the problem is somewhere in `vanchorWithdraw()` inside `vanchorPrivateTransaction.test.ts`. That file had some hardcoded values before which I had to replace, and most likely there is something wrong.

drewstone:
What's the status of this @Nutomic ?

Nutomic:
I cleaned up the Rust code now, adding error handling, fixing todos and making clippy pass. So only thing left is fixing the tests.

crates/relayer-handler-utils/src/lib.rs
```diff
@@ -171,7 +171,7 @@ type R = Vec<[u8; 32]>; // Substrate roots format
 type E = [u8; 32]; // Substrate element type
 type I = AccountId32; // Substrate account identifier
 type B = U128; // Substrate balance type
-type A = i128; // Substrate signed amount type
+type A = WebbI256; // Substrate signed amount type
```
Nutomic, line 174: Is this correct? Based on the original code it should be i128. Although thats causing errors so we might have to create `WebbI128`.

shekohex: Yes please, create a custom type here like `WebbI128`

services/webb-relayer/src/service/evm.rs
```diff
@@ -52,8 +52,8 @@ pub fn build_web_services() -> Router<Arc<RelayerContext>> {
         // for backward compatibility
         .route("/metrics", get(metric::handle_metric_info))
         .route(
-            "/fee_info/:chain_id/:vanchor/:gas_amount",
-            get(handle_fee_info),
+            "/fee_info/evm/:chain_id/:vanchor/:gas_amount",
+            get(handle_evm_fee_info),
```
shekohex, line 56: This a breaking change, so we may want to outline this into the PR overview?

shekohex: Will be useful for the dApp team to do these updates too.

shekohex (changes requested):
Looks good! just few things needs that needs more clean-ups

services/webb-relayer/src/service/evm.rs
shekohex: cc @AtelyPham @dutterbutter

Nutomic: I mentioned it in the summary of changes, should I add anything there?

crates/tx-relay/src/evm/vanchor.rs
```diff
@@ -81,19 +82,6 @@ pub async fn handle_vanchor_relay_tx<'a>(
...
-        .await
-        .map_err(|e| {
-            Error(format!("Failed to retrieve relayer balance: {e}"))
-        })?;
-    if cmd.ext_data.refund > relayer_balance {
-        return Err(Error(
-            "Requested refund is higher than relayer balance".to_string(),
-        ));
-    }
-
```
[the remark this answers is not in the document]
salman01zp: This check was removed cc: @Nutomic  adding it back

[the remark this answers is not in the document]
Nutomic: Line 159 (because fee_info was changed to check the balance when calculating max_refund).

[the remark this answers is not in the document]
salman01zp: I see..

services/webb-relayer/src/service/evm.rs
shekohex: No, all is good.

crates/tx-relay/src/evm/vanchor.rs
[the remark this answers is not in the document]
shekohex: Relayer balance could change from the moment where we cached the fees and when we executing the actual tx.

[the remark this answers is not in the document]
Nutomic: Okay will add code to recheck the balance when using cached value.

shekohex (approved):
Awesome work everyone, LGTM!

PR #407, opened 2023-03-29
Add `assets` to the Relayer config

### LINKED ISSUE #404: [BUG] Relayer panics with `no entry found for key` while fetching price of the wrapped token
### Reproduce the issue

Using the current deployed Relayer 4
```bash
curl -sw "%{http_code}\n" https://relayer4.webb.tools/api/v1/fee_info/1099511627781/0x9678647b9fcb0039652a16dba688bd067d6e5077/0
```
You will see the result status code is `502`

Looking into relayer logs on the server we find the following:

```log
thread 'tokio-runtime-worker' panicked at 'no entry found for key', /project/crates/tx-relay/src/evm/fees.rs:134:31
stack backtrace:
    0:     0x7f75dce9b16e - <unknown>
    1:     0x7f75dc99656e - <unknown>
    2:     0x7f75dce97c75 - <unknown>
    3:     0x7f75dce9af45 - <unknown>
    4:     0x7f75dce9c61f - <unknown>
....
```

Here is the relative code:

https://github.com/webb-tools/relayer/blob/02a1ea9a804ac68abfe8e793a9a8a54420fa57d3/crates/tx-relay/src/evm/fees.rs#L133-L134

### LINKED ISSUE #406: [SPEC] Relayer Open-Market for unlisted Assets
### Overview

In the world of cryptocurrency, there are many assets available for trading. However, not all of them are listed on popular exchanges, which makes it difficult for us to fetch the prices of that asset. An open-market for unlisted crypto assets can be very useful in this case.
An open-market for unlisted assets is a feature where the Relayer operator can decide the price of an asset using it's symbol (`tTNT`, `WEBB`, ..etc). The price is determined based on the demand and supply of the asset in question, without the need for bidding or an open-close bookkeeping system, It is really up to the Relayer operator to set the price for any asset, and the user is free to choose which relayer to use and see it fits its needs. A good and honest UI would *optionally* query all the available relayers to get a good quote of the transaction, and sort them by the best price & fees.

### Proposed Config Changes

in the config files we can add a new config option under the global section for asset prices.

For example,

```toml
# Assets in assets.toml
[assets]
tTNT.price = 0.10 # 10 cents
TNT.price = 0.1
# ..etc
```

Also, we should make sure that the asset prices appears when we query the `api/v1/info` endpoint, like the following:

```json
{
  "evm": {},
  "substrate": {},
  "experimental": {},
  "features": {},
  "assets": {
    "tTNT": {
      "price": 0.1
    },
    "TNT": {
      "price": 0.1
    }
  }
}
```

### Using the Configured Asset prices

While querying the Price Oracle API for a price of a cretin token, and we can't find this said token symbol listed on that Oracle, we fallback to read it from our configured assets, However, if we can't find it there either, we should return an error to the user saying `Unsupported  Asset` for example.

### SPEC Checklist

- [ ] Add Support for the `assets` section in the `WebbRelayerConfig`
- [ ] Make sure it is an optional field, defaults to an our pre-configured prices (see below).
- [ ] Change the logic for query the token prices, according to the above instructions.
- [ ] Update the `config/README.md` to add that section to the docs.
- [ ] Update tests if needed, like the typescript configuration. 

### Default Assets

```rust
fn default_assets() -> HashMap<String, UnlistedAsset> {
    HashMap::from_iter([
       (String::from("tTNT"), UnlistedAsset { price: 0.1 }), 
       (String::from("TNT"), UnlistedAsset { price: 0.1 }), 
   ])
}
```

### Related Issues

#404

shekohex:
## Summary of changes
- Adds the `assets` section to the config
- Implements the SPEC in #406 
- Fixes the bug where the relayer would panic if the token is not found on the price oracle.

- Closes #404 
- Closes #406 

-----
### Code Checklist 

- [x] Tested
- [x] Documented


salman01zp (commented):
Looks good.. Just need CI to pass.

salman01zp (approved):

crates/tx-relay/src/evm/fees.rs
```diff
@@ -114,25 +113,46 @@ async fn generate_fee_info(
...
             false,
             false,
             false,
             false,
         )
         .await?;
-    let native_token_price = prices[native_token.0].usd.unwrap();
-    let wrapped_token_price = prices[&wrapped_token.0].usd.unwrap();
+    let native_token_price = match prices.get(native_token) {
+        Some(price) => price.usd.expect("price.usd is not None"),
```
Nutomic, line 134: expect panics if there is None. Would be better to handle this gracefully by returning an error. I would rewrite this whole match statement like `prices.get(native_token).map(|t| t.price).ok_or(|| /* return error here */)`

crates/tx-relay/src/evm/fees.rs
```diff
@@ -114,25 +113,46 @@ async fn generate_fee_info(
...
+        .or(ctx.config.assets.get(&wrapped_token).map(|a| a.price));
 
+    let wrapped_token_price = match maybe_wrapped_token_price {
+        Some(price) => price,
+        None => {
+            return Err(webb_relayer_utils::Error::FetchTokenPriceError {
+                token: wrapped_token.clone(),
+            })
+        }
+    };
```
Nutomic, line 155: Same as above, use `maybe_wrapped_token_price.map(|t| t.price).ok_or()` then there is no need for variable maybe_wrapped_token_price.

PR #408, opened 2023-04-03
Add more clippy checks to avoid potential runtime crashes

### LINKED ISSUE #383: [TASK] Add more clippy checks to the CI to avoid runtime crashes
### Overview

In our effort to make our relayer system more stable, we are working to detect and remove any possible runtime crashes due to the use of `unwrap` or `expect` or any other means of panicking during the runtime instead of just handling the error gracefully.

### Task Checklist

- [x] Add Clippy Checks to the CI similar to this one [here](https://github.com/LemmyNet/lemmy/blob/fdaa2bcf090c3fb1c2c38fe7bac8026ce67457d3/scripts/fix-clippy.sh#L8-L21)
- [x] Make sure that CI creates the warning and comments on the PR by using https://github.com/actions-rs/clippy-check
- [x] Update and Fix any clippy warnings in the same PR.

Nutomic:
## Summary of changes
- Get rid of unwrap usage
- Return Err in case substrate storage returns None value
- Apply various other lints to improve code quality and readability

- Closes #383

-----
### Code Checklist 

- [ ] Tested
- [ ] Documented

### DIFF
--- a/crates/relayer-handlers/src/lib.rs
+++ b/crates/relayer-handlers/src/lib.rs
@@ -39,10 +39,10 @@ use webb_relayer_handler_utils::{
 };
 use webb_relayer_tx_relay::evm::fees::{get_fee_info, FeeInfo};
 
-use crate::routes::HandlerError;
 use webb_relayer_tx_relay::evm::vanchor::handle_vanchor_relay_tx;
 use webb_relayer_tx_relay::substrate::mixer::handle_substrate_mixer_relay_tx;
 use webb_relayer_tx_relay::substrate::vanchor::handle_substrate_vanchor_relay_tx;
+use webb_relayer_utils::HandlerError;
 
 /// Module handles relayer API
 pub mod routes;
--- a/crates/relayer-handlers/src/lib.rs
+++ b/crates/relayer-handlers/src/lib.rs
@@ -17,7 +17,6 @@
 #![allow(clippy::large_enum_variant)]
 #![warn(missing_docs)]
 use axum::extract::{Path, State, WebSocketUpgrade};
-use axum::http::StatusCode;
 use ethereum_types::{Address, U256};
 use std::error::Error;
 use std::sync::Arc;
@@ -208,19 +207,7 @@ pub async fn handle_fee_info(
 ) -> Result<Json<FeeInfo>, HandlerError> {
     let chain_id = TypedChainId::from(chain_id);
     let gas_amount = U256::from(gas_amount);
-    get_fee_info(chain_id, vanchor, gas_amount, ctx.as_ref())
+    Ok(get_fee_info(chain_id, vanchor, gas_amount, ctx.as_ref())
         .await
-        .map(Json)
-        .map_err(|e| {
-            let status = match e {
-                webb_relayer_utils::Error::FetchTokenPriceError { .. } => {
-                    StatusCode::BAD_REQUEST
-                }
-                webb_relayer_utils::Error::EtherscanConfigNotFound {
-                    ..
-                } => StatusCode::BAD_REQUEST,
-                _ => StatusCode::INTERNAL_SERVER_ERROR,
-            };
-            HandlerError(status, e.to_string())
-        })
+        .map(Json)?)
 }
--- a/crates/tx-relay/src/evm/vanchor.rs
+++ b/crates/tx-relay/src/evm/vanchor.rs
@@ -1,4 +1,4 @@
-use super::*;
+use super::{CommandResponse, Middleware, WithdrawStatus};
 use crate::evm::fees::{get_fee_info, FeeInfo};
 use crate::evm::handle_evm_tx;
 use ethereum_types::U256;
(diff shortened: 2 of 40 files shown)


event-watchers/substrate/src/vanchor_encrypted_output_handler.rs
```diff
@@ -68,7 +68,7 @@ impl EventHandler<SubstrateConfig> for SubstrateVAnchorEncryptedOutputHandler {
                 .await?
                 .fetch(&next_leaf_index_addr)
                 .await?
-                .unwrap();
+                .expect("fetch next leaf index from storage");
```
Nutomic, line 71: Not sure whats the best way to handle this, maybe return an Err?

shekohex: Actually, whenever we fail to fetch a storage value, for whatever reason, we should return with an error. Please do not use `unwrap_or_default` for this case, the default case here is not desirable.

shekohex:
Is this PR ready for reviews @Nutomic ?

event-watchers/substrate/src/signature_bridge_watcher.rs
```diff
@@ -352,16 +352,6 @@ where
...
 
-pub fn parse_nonce_from_proposal_data(proposal_data: &[u8]) -> u32 {
-    let nonce_bytes = proposal_data[36..40].try_into().unwrap_or_default();
-    u32::from_be_bytes(nonce_bytes)
-}
-
-pub fn parse_call_from_proposal_data(proposal_data: &[u8]) -> Vec<u8> {
-    // Not [36..] because there are 4 byte of zero padding to match Solidity side
-    proposal_data[40..].to_vec()
-}
```
Nutomic, line 363: Unused code. Sadly there seems to be no working tool which can find unused code across a cargo workspace.

crates/tx-relay/src/evm/vanchor.rs
```diff
@@ -122,8 +122,16 @@ pub async fn handle_vanchor_relay_tx<'a>(
...
+            cmd.proof_data
+                .output_commitments
+                .get(0)
+                .map(|c| c.to_fixed_bytes().into())
+                .unwrap_or_default(),
+            cmd.proof_data
+                .output_commitments
+                .get(1)
+                .map(|c| c.to_fixed_bytes().into())
+                .unwrap_or_default(),
```
Nutomic, line 134: In case of missing values these should return default U256 with all zeros, which will fail the signature check.

crates/relayer-handlers/src/lib.rs
```diff
@@ -208,19 +207,7 @@ pub async fn handle_fee_info(
...
-                    StatusCode::BAD_REQUEST
-                }
-                webb_relayer_utils::Error::EtherscanConfigNotFound {
-                    ..
-                } => StatusCode::BAD_REQUEST,
-                _ => StatusCode::INTERNAL_SERVER_ERROR,
-            };
-            HandlerError(status, e.to_string())
-        })
+        .map(Json)?)
```
Nutomic, line 212: Like I commented before, none of these are actually client errors, so status code internal server error is more appropriate.

Nutomic:
From my side the PR is finished now.

crates/tx-relay/src/evm/vanchor.rs
shekohex: Instead of doing that, could we just map all the `cmd.proof_data.output_commitments` into the expected output commitments as bytes? Like:

```suggestion
                cmd.proof_data
                .output_commitments
                .into_iter()
                .map(|c| c.to_fixed_bytes().into())
                .collect()
```

That would be useful in the future to also support more than two commitments.

event-watchers/evm/src/vanchor/vanchor_leaves_handler.rs
```diff
@@ -43,19 +43,25 @@ pub struct VAnchorLeavesHandler {
...
+    ) -> webb_relayer_utils::Result<Self> {
         let params = setup_params::<Bn254Fr>(Curve::Bn254, 5, 3);
         let poseidon = Poseidon::<Bn254Fr>::new(params);
         let default_leaf_scalar: Vec<Bn254Fr> =
             bytes_vec_to_f(&vec![default_leaf]);
-        let default_leaf_vec = default_leaf_scalar[0].into_repr().to_bytes_be();
+        let default_leaf_vec = default_leaf_scalar
+            .get(0)
+            .map(|d| d.into_repr().to_bytes_be())
+            .unwrap_or_default();
```
shekohex, line 57: This one also should return an Error that say we can't convert the default leaf scalar into bytes.

crates/tx-relay/src/evm/vanchor.rs
Nutomic: Had to adjust it a bit but it works.

event-watchers/evm/src/vanchor/vanchor_leaves_handler.rs
Nutomic: Done

shekohex (approved):
Looks good!

PR #411, opened 2023-04-03
[feat] Introduce Price Oracle Backends

### LINKED ISSUE #402: [BUG]  Coingecko API is down, need better solution for fetching token prices
## Summary
Currently, we just use coin gecko API to fetch prices for tokens, but it went down and due to this issue, the application is unable to fetch real-time cryptocurrency prices and fails with an error
```
reqwest::Error { kind: Decode, source: Error(\"expected value\", line: 1, column: 1) }"
```

### Solution 1
  Use alternative API like coinbase if coingecko API is down.

### Solution 2
   Use uniswap oracle contract to fetch token prices

shekohex:
## Summary of changes
- Added Price Oracle Backend trait
- Implemented Coingecko as a backend
- Added a CachedPriceOracle backend that would work with any price oracle
- Added a DummyPriceOracle backend that would work with hardcoded prices (e.g for unlisted tokens)
- Added a PriceOracleMerger backend that merges more than one price oracle backend(s) into one!
- Update the hardcoded chain ids (should be resolved and will no need for hardcoding them after #412 )

- Closes #402 

-----
### Code Checklist 

- [x] Tested
- [x] Documented

### DIFF
--- a/crates/price-oracle-backends/src/cached.rs
+++ b/crates/price-oracle-backends/src/cached.rs
@@ -16,20 +16,33 @@ pub struct CachedPriceBackend<B, S> {
     backend: B,
     /// The local data store used for caching
     store: S,
-    /// The cache expiration time in seconds
+    /// The cache expiration time.
+    ///
     /// If the cache is older than this value, it will be refreshed
+    /// from the source backend.
+    ///
     /// If the value is `None`, the cache will never expire
-    /// The default value is `15 minutes`.
+    /// and will never be refreshed. **This may lead to incorrect price data.**
+    ///
+    /// Use this option only if you are sure that the source backend is always available.
+    /// Otherwise, use a reasonable value, The default value is `15 minutes`.
+    /// see [`Self::use_cache_if_source_unavailable`] and [`Self::even_if_expired`]
+    /// for fine tuning the cache behavior.
     #[builder(default = Some(Duration::from_secs(15 * 60)))]
     cache_expiration: Option<Duration>,
     /// Specifies whether the cache should be returned even if the source is unavailable
     ///
     /// If the value is `true`, the cache will be returned even if the source
-    /// backend is unavailable even if the cache is expired.
+    /// backend is unavailable unless the cache is expired.
+    ///
+    /// see [`Self::even_if_expired`] if you want to return the cache even if it is expired.
     #[builder(setter(strip_bool))]
     use_cache_if_source_unavailable: bool,
     /// Specifies whether the cache should be returned even if it is expired
     /// in case the source is unavailable.
+    ///
+    /// see [`Self::use_cache_if_source_unavailable`] if you want to return the cache
+    /// even if the source is unavailable.
     #[builder(setter(strip_bool))]
     even_if_expired: bool,
 }
--- a/crates/price-oracle-backends/src/cached.rs
+++ b/crates/price-oracle-backends/src/cached.rs
@@ -28,6 +28,10 @@ pub struct CachedPriceBackend<B, S> {
     /// backend is unavailable even if the cache is expired.
     #[builder(setter(strip_bool))]
     use_cache_if_source_unavailable: bool,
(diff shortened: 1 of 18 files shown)


crates/price-oracle-backends/src/lib.rs
```diff
@@ -0,0 +1,274 @@
...
+//! ## With Cache
+//! ```rust,no_run
+//! use price_oracle_backends::{CoinGeckoBackend, PriceBackend, CachedPriceBackend};
+//! use webb_relayer_store::SledStore;
+//! let store = SledStore::temporary()?;
+//! let backend = CoinGeckoBackend::new();
+//! let backend = CachedPriceBackend::new(backend, store);
+//! let prices = backend.get_prices(&["ETH", "BTC"]).await?;
+//! let eth_price = prices.get("ETH").expect("ETH price is missing");
+//! let btc_price = prices.get("BTC").expect("BTC price is missing");
```
Nutomic, line 44: In tests I would always use unwrap, its easier to read and failures are easy to find compared to normal code.

crates/price-oracle-backends/src/lib.rs
```diff
@@ -0,0 +1,274 @@
...
+pub struct CachedPriceBackend<B, S> {
+    /// The price backend
+    backend: B,
+    /// The local data store used for caching
+    store: S,
+    /// The cache expiration time in seconds
+    /// If the cache is older than this value, it will be refreshed
+    /// If the value is `None`, the cache will never expire
+    /// The default value is `15 minutes`.
+    #[builder(default = Some(Duration::from_secs(15 * 60)), setter(strip_option))]
```
Nutomic, line 149: There should be a method `Duration::from_mins`

shekohex: I wish, but there is no `Duration::from_mins` method.

crates/price-oracle-backends/src/cached.rs
```diff
@@ -0,0 +1,210 @@
...
+            }
+        }
+        if !tokens_to_fetch.is_empty() {
+            // Fetch the prices from the source
+            let token_ids = tokens_to_fetch.iter().copied().collect::<Vec<_>>();
+            let result = self
+                .backend
+                .get_prices_vs_currency(&token_ids, vs_currency)
+                .await;
+            let soruce_unavailable = result.is_err();
```
drewstone, line 133:
```suggestion
            let source_unavailable = result.is_err();
```

crates/price-oracle-backends/src/cached.rs
```diff
@@ -0,0 +1,210 @@
...
+                        super::PricesMap::new()
+                    } else {
+                        return Err(err);
+                    }
+                }
+            };
+
+            // If the source is unavailable and the cache is enabled and `even_if_expired` is enabled,
+            // return the cache
+            if soruce_unavailable
```
drewstone, line 148:
```suggestion
            if source_unavailable
```

crates/price-oracle-backends/src/cached.rs
```diff
@@ -0,0 +1,210 @@
...
+                for token in tokens {
+                    let token_key = format!("{token}/{vs_currency}");
+                    if let Some(cached) = self.store.get_price(&token_key)? {
+                        prices.insert((*token).to_owned(), cached.price);
+                    }
+                }
+            }
+
+            // Update the cache, only if the source is available
+            let soruce_available = !soruce_unavailable;
```
drewstone, line 162:
```suggestion
            let source_available = !source_unavailable;
```

crates/price-oracle-backends/src/cached.rs
```diff
@@ -0,0 +1,210 @@
...
+            if soruce_available {
```
drewstone, line 163:
```suggestion
            if source_available {
```

drewstone (changes requested):
Spelling

Issue #437, opened 2023-04-06
[TASK] Batch Reward Tree Relaying: Proof Generation


saiakilesh:
- [ ]  Function that takes in a set of leaves (and any additional args) and generates a batch reward unspent/spent tree proof
    - [ ]  Leaves are reward unspent tree commitments
        - [ ]  Insert leaves into Merkle tree and get new root of Merkle root.
        - [ ]  Take in currentRoot, newRoot, and leaves and create witness.
        - [ ]  Tests:
            - [ ]  Success test: generated proof verifies with 4 insertions
            - [ ]  Negative test: proof generated cannot be generated for wrong Merkle tree root

Issue #434, opened 2023-04-06
[TASK] Batch Deposit Relaying: Event Listener


akileshtangella:
- [ ]  Event listener for the batched deposit queue.
    - [ ]  Create autogen code in Rust for proxy contract in `webb-rs` repo. Add code to `build.rs`.
    - [ ]  Create an event listener for MASP proxy. Leverage code from other event listeners.
    - [ ]  Create an event handler that on a newly queued deposit, check if number of queued deposits mod 4 is 0 (otherwise continue.
    - [ ]  If it is, create a batch update proof and submit it back to proxy.

PR #453, opened 2023-04-10
Masp Batch Deposits

semaraugusto:
## Summary of changes
- Adds proof generation for the batch deposit functionality on the MASP
- Adds `masp_leaves_handler`, 
- Adds `masp_proxy_handler`

- Closes https://github.com/webb-tools/relayer/issues/438
- Closes https://github.com/webb-tools/relayer/issues/434
- Closes https://github.com/webb-tools/relayer/issues/437 (proof doesnt depend on which tree we're using)

-----
### Code Checklist 

- [ ] Tested
- [ ] Documented


salman01zp:
Can you also link SPEC for MASP Relayer Integration to this PR..?

semaraugusto:
base relayer integrations spec: https://www.notion.so/MASP-Relayer-Integration-8d4f6e4c8e124dc1b87bac4766783bd5

batch deposit specific spec: https://www.notion.so/hicommonwealth/Batched-Reward-Tree-Processing-7d51b34188584288af9266e7f2a06542

crates/relayer-config/src/evm/mod.rs
```diff
@@ -120,6 +120,39 @@ pub struct VAnchorContractConfig {
...
+#[serde(rename_all(serialize = "camelCase", deserialize = "kebab-case"))]
+pub struct MaspProxyContractConfig {
+    /// Common contract configuration.
+    #[serde(flatten)]
+    pub common: CommonContractConfig,
+    /// Controls the events watcher
+    pub events_watcher: EventsWatcherConfig,
+    /// The type of the optional signing backend used for signing proposals. It can be None for pure Tx relayers
+    #[serde(rename(serialize = "proposalSigningBackend"))]
+    pub proposal_signing_backend: Option<ProposalSigningBackendConfig>,
```
drewstone, line 151: Is this needed or used anywhere on the proxy config? I recommend starting with only the things we use so these objects don't get clogged.

event-watchers/evm/src/masp/masp_generate_batch_deposit_proof.rs
```diff
@@ -0,0 +1,122 @@
...
+    use super::*;
+
+    #[test]
+    fn test_gen_proof() {
+        let zkey_path =
+            "../../tests/solidity-fixtures/batch-tree/4/circuit_final.zkey";
+        let wasm_path = "../../tests/solidity-fixtures/batch-tree/4/batchMerkleTreeUpdate_4.wasm";
+        let prover = MaspBatchProver::new(zkey_path, wasm_path);
+
+        // pre-generated inputs
```
drewstone, line 73: Please instruct how to generate these pre-generated inputs too. All pre-generated fixture tests should be reproducible from the first step.

event-watchers/evm/src/masp/masp_leaves_handler.rs
```diff
@@ -0,0 +1,228 @@
...
+                let poseidon = Poseidon::<Bn254Fr>::new(params);
+                batch.insert(leaf_index, leaf);
+                let mut mt = self.mt.lock().await;
+                mt.insert_batch(&batch, &poseidon)?;
+                let root_bytes = mt.root().into_repr().to_bytes_be();
+                let root: U256 = U256::from_big_endian(root_bytes.as_slice());
+                let is_known_root =
+                    wrapper.contract.is_known_root(root).call().await?;
+                tracing::trace!("Is known root: {:?}", is_known_root);
+                if event_data.leaf_index.as_u32() % 2 == 0 {
```
drewstone, line 101: If we are requiring the local MT for the relayer to be in sync, wouldn't this prevent that if the leaf index is even? Should this be removed?

drewstone: Is this function copied from the VAnchor leaf watcher? If so, are we duplicating code that can otherwise be consolidated?

drewstone (changes requested):

event-watchers/evm/src/masp/masp_generate_batch_deposit_proof.rs
```diff
@@ -0,0 +1,122 @@
+use ark_bn254::{Bn254, Fr};
```
drewstone, line 1: Thoughts on moving files like these into a separate crate? Something dedicated to all our proof generation needs?

drewstone (changes requested):

Issue #449, opened 2023-04-10
[CHECKLIST] Automate The Relayer Deployment


shekohex:
### Overview

We aim to achieve a ~%100 deployment automation here, from the PR gets merged in the develop branch to the creating a new release and finally deployed on the server(s).

### The Automation Flow

**Development deployment flow** (`develop` branch)

1. Opening a PR to fix a bug in the `develop` branch.
2. CI Starts to check the code that it **only** compiles on Linux targets (no lints, no integration tests, ..etc).
3. Once reviewed, it gets merged _automatically_ , that's it.
4. Creating a new release is automatic, by just creating a new tag. (should take less than 30s to do so)
5. CI notices a new release, it builds a new docker image and publish it to that release tag.
6. After that, the same CI, zips the current deployment config files and adds it to that release.
7. Once Above CI Action finishes, another one starts up to SSH into the Relayer development* boxes to deploy the new changes by running a script we added on the server

Target time for all CI steps from opening a PR: **~10min**.
Target time for all CI steps from creating a new release tag: **~20min**.

> Development Relayers are {Relayer4, Relayer5}

**Production deployment flow** (`main` branch)

1. We open a PR to merge `develop` into `main` branch.
7. CI Starts to Run (lints, unit-tests, integration-tests, ..etc)
8. If any of the checks failed, we open a clean up PR(s) to address these issues and merge it into `develop`
9. Once the CI is green on `main` we double-check everything and then we merge it (manually)
10. Create a new **stable** tag (manually)
11. The same CI from above notices a new **stable** release got published, it builds a new docker image and publishes it to that release tag.
12. After that, it publishes the current deployment config files to that release.
13. Once done, the other CI notices the new **stable** release and SSH into the Relayer production* boxes to deploy the new stable relayer by running a script we added on the server.

> Production Relayers are {Relayer1, Relayer2, Relayer3}

Target time for all CI steps from opening a PR: **~1hour**.
Target time for all CI steps from creating a new release tag: **~20min**.

### Checklist Tasks

- [ ] Update the Github Actions files to be in the following structure
    - `lints.yml` For Clippy & Rust format
    - `build.yml` For building the relayer on (Linux, macOS)
    - `tests.yml` For Running the relayer tests on Linux.
    - `image.yml` For building and publishing the Docker image + config files
    - `publish-rust-docs.yml` For publishing the docs.
    - `deploy.yml` for running the deployment scripts.
[...]

PR #454, opened 2023-04-11
Attempt to speedup CI steps (ref #449)

Nutomic:
## Summary of changes
- Remove unnecessary cargo build step for integration tests
- Use `cargo check` instead of `cargo build`
- Dont run tests in release mode

This reduces the total runtime of tests from [3h 3m](https://github.com/webb-tools/relayer/actions/runs/4657577954/usage) to [1h 41m](https://github.com/webb-tools/relayer/actions/runs/4666652799/usage). As far as I can tell the checks are equivalent in both cases. However it still takes roughly 30m like before to finish all tests, because `linux-integration-tests` is quite slow. We could try to improve this by splitting integration tests in two, then evm and substrate tests could run in parallel which should roughly cut the total time in half.

- Closes 

-----
### Code Checklist 

- [ ] Tested
- [ ] Documented

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -17,7 +17,7 @@ env:
 
 jobs:
 
-  macos-build:
+  macos-check:
     runs-on: macos-latest
     steps:
       - name: Cancel Previous Runs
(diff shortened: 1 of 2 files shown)


shekohex (approved):
Looks good!

PR #458, opened 2023-04-12
Run CI integration tests in parallel

### LINKED ISSUE #455: [TASK] Speed up CI
The CI is relatively slow with around 30 minutes to finish. We should find ways to speed it up. I already made some progress on this in https://github.com/webb-tools/relayer/pull/454 which reduces the total runtime of tests from [3h 3m](https://github.com/webb-tools/relayer/actions/runs/4657577954/usage) to [1h 41m](https://github.com/webb-tools/relayer/actions/runs/4666652799/usage). 

However we still need to wait around 30 minutes for integration tests, so those need to be faster. One solution worth trying is to run tests in parallel, although this is unlikely to work. Another possibility is to split the integration tests in two for CI, so that evm and substrate tests can be run in parallel.

Nutomic:
## Summary of changes
- Execute integration tests for evm/substrate in separate tasks  so that they can run in parallel. 

- Closes #455

-----
### Code Checklist 

- [ ] Tested
- [ ] Documented


Nutomic:
This is odd, evm integration tests are failing because of an assertion failure in `Vanchor Transaction relayer`, and for some reason substrate integration tests are cancelled automatically before starting every time. Not sure what the problem might be, maybe there is some hidden dependency between different tests?

shekohex (approved):

shekohex:
> This is odd, evm integration tests are failing because of an assertion failure in `Vanchor Transaction relayer`, and for some reason substrate integration tests are cancelled automatically before starting every time. Not sure what the problem might be, maybe there is some hidden dependency between different tests?

I guess it is because of the [`fail-fast`](https://github.com/webb-tools/relayer/blob/12194686c868cb1b5d03964ee5fa193140271f30/.github/workflows/check.yml#L117)

Nutomic:
After a couple of restarts everything is passing. Time went down from 30m to 20m. One test is still listed as pending because I removed it (it wasnt doing anything). Not sure how to get rid of that one. 

> I guess it is because of the [fail-fast](https://github.com/webb-tools/relayer/blob/12194686c868cb1b5d03964ee5fa193140271f30/.github/workflows/check.yml#L117)

Ah makes sense. Still weird that evm tests failed.

shekohex:
Updated my comment with the right one. For that I only got the sense that it is not gonna be feasible from your comments. feel free to re-open this PR if you think otherwise.

Nutomic:
Okay reopened. I think it should work fine if you or a repo admin removes `linux-integration-tests (stable, aarch64-unknown-linux-musl)` from https://github.com/webb-tools/relayer/settings/branch_protection_rules/

salman01zp:
We can remove `linux-integration-tests (stable, aarch64-unknown-linux-musl)` from the mandatory test for develop branch.
@shekohex @dutterbutter

salman01zp (approved):
LGTM ..!!

shekohex:
> We can remove `linux-integration-tests (stable, aarch64-unknown-linux-musl)` from the mandatory test for develop branch. @shekohex @dutterbutter

These are the only required checks in `develop` branch:
image

Nutomic:
This PR removes the check `linux-integration-tests (stable, aarch64-unknown-linux-musl)` and adds two separate ones instead:
- `linux-integration-tests (stable, x86_64-unknown-linux-musl, test-evm)`
- `linux-integration-tests (stable, x86_64-unknown-linux-musl, test-substrate)`

This means that the same tests are run, but on two separate machines so that it can be in parallel. Merging this PR requires removing `linux-integration-tests (stable, aarch64-unknown-linux-musl)` from mandatory checks and adding the two new ones instead.

PR #461, opened 2023-04-14
Use transaction queue for DKG/Substrate (fixes #450)

### LINKED ISSUE #450: [TASK] Process DKG/Substrate extrinsic calls  through Tx queue
## Overview

DKG should make use of the Substrate Transaction Queue instead of in-place sending transactions.

### Task Checklist

- [ ] Create transactions in-place and encode them.
- [ ] Make use of the Substrate tx queue.
- [ ] Process all extrinsic calls made to DKG and substrate through the tx queue like from [here](https://github.com/webb-tools/relayer/blob/main/crates/proposal-signing-backends/src/dkg.rs#L113)

Nutomic:
## Summary of changes
- Use substrate transaction queue instead of sending in place

- Closes #450

-----
### Code Checklist 

- [ ] Tested
- [ ] Documented


crates/proposal-signing-backends/src/dkg.rs
```diff
@@ -87,82 +100,32 @@ impl super::ProposalSigningBackend for DkgProposalSigningBackend {
...
-                }
-                TransactionStatus::Usurped(_) => {}
-                TransactionStatus::Dropped => {
-                    tracing::warn!("tx dropped");
-                }
-                TransactionStatus::Invalid => {
-                    tracing::warn!("tx invalid");
-                }
-            }
-        }
```
Nutomic, line 165: There is no progress returned so this code doesnt work anymore. I assume similar logs are already written somewhere in the tx queue.

crates/proposal-signing-backends/src/dkg.rs
```diff
@@ -87,82 +100,28 @@ impl super::ProposalSigningBackend for DkgProposalSigningBackend {
...
-                    tracing::error!(error = %err, "failed to watch for tx events");
-                    return Err(err.into());
-                }
-            };
+        let execute_proposal_call = ExecuteProposal {
+            src_id: self.typed_chain_id.chain_id(),
+            proposal_data: BoundedVec(proposal.to_vec()),
+            signature: BoundedVec(signature.encode()),
+        };
+        enqueue_transaction(execute_proposal_call, self.store.as_ref())?;
```
shekohex, line 123: Well first of all, this does not sound right to me, the original code used the `acknowledge_proposal` call, not `execute_proposal` call, which are two different extrinsic from two different pallets. 

Besides, there is no need for signing anything, why you used the `self.pair.sign` call here? the TxQueue is responsible for signing proposals.

crates/proposal-signing-backends/src/dkg.rs
```diff
@@ -190,3 +149,45 @@ fn webb_proposals_typed_chain_converter(
...
+    QueueStore::<WebbDynamicTxPayload>::enqueue_item(
+        store,
+        tx_key,
+        execute_proposal_tx,
+    )?;
+    tracing::debug!(
+        data_hash = ?hex::encode(data_hash),
+        "Enqueued execute-proposal call for execution through protocol-substrate tx queue",
+    );
+    Ok(())
```
shekohex, line 184: There is no need for creating a function here, each transaction has to be constructed on its own, you are creating a function to enqueue the wrong transaction, not even the one originally in the code. Please update the code to use the original transaction.

shekohex (changes requested):
Not entirely there yet, but looks a like a good start.

crates/proposal-signing-backends/src/dkg.rs
Nutomic: Okay removed this.

crates/proposal-signing-backends/src/dkg.rs
Nutomic: It wasnt clear to me which extrinsic it is using here, fixed that. The queue docs also dont mention anything about signing, so I assumed that it has to be passed exactly like in signature bridge watcher. Should be fixed now.

crates/proposal-signing-backends/src/dkg.rs
```diff
@@ -85,7 +95,7 @@ impl super::ProposalSigningBackend for DkgProposalSigningBackend {
     async fn handle_proposal(
         &self,
         proposal: &(impl ProposalTrait + Sync + Send + 'static),
-        metrics: Arc<Mutex<metric::Metrics>>,
+        _metrics: Arc<Mutex<metric::Metrics>>,
```
Nutomic, line 98: Metrics are unused now, not sure if we should use it for something else or remove the param.

shekohex: We would use it in the future, I assume by adding more metrics to the system.

crates/proposal-signing-backends/src/dkg.rs
```diff
@@ -96,73 +106,38 @@ impl super::ProposalSigningBackend for DkgProposalSigningBackend {
...
         );
+        // webb dynamic payload
+        let execute_proposal_tx = WebbDynamicTxPayload {
+            pallet_name: Cow::Borrowed("dkgProposals"),
+            call_name: Cow::Borrowed("acknowledge_proposal"),
+            fields: vec![
+                Value::u128(u128::from(self.typed_chain_id.chain_id())),
+                Value::from_bytes(xt.call_data().encode()),
+            ],
+        };
```
shekohex, line 128: This call is wrong, first of all the pallet name is `DKGProposals` as you can find it here in the runtime:
https://docs.rs/webb/0.5.21/src/webb/substrate/dkg_runtime.rs.html#14

Another this is the parameters does not look right to me, this call takes 4 args (nonce, src_chain_id, resource_id, proposal_bytes)
Here is a suggested fix:
```suggestion
        let execute_proposal_tx = WebbDynamicTxPayload {
            pallet_name: Cow::Borrowed("DKGProposals"),
            call_name: Cow::Borrowed("acknowledge_proposal"),
            fields: vec![
                Value::u128(u128::from(nonce)),
                Value::u128(u128::from(self.typed_chain_id.chain_id())),
                Value::from_bytes(ResourceId(resource_id.into_bytes()).encode()),
                Value::from_bytes(BoundedVec(proposal.to_vec()).encode()),
            ],
        };
```

crates/proposal-signing-backends/src/dkg.rs
```diff
@@ -96,73 +106,38 @@ impl super::ProposalSigningBackend for DkgProposalSigningBackend {
...
+        let data_hash = utils::keccak256(xt.call_data().encode());
+        let tx_key = SledQueueKey::from_substrate_with_custom_key(
+            self.typed_chain_id.underlying_chain_id(),
+            make_execute_proposal_key(data_hash),
```
shekohex, line 132: function name here is misleading, there is not execute proposal in the process, I would suggest renaming it to something like:
```suggestion
            make_ acknowledge_proposal_key(data_hash),
```

crates/proposal-signing-backends/src/dkg.rs
```diff
@@ -190,3 +165,11 @@ fn webb_proposals_typed_chain_converter(
...
     }
 }
+
+pub fn make_execute_proposal_key(data_hash: [u8; 32]) -> [u8; 64] {
+    let mut result = [0u8; 64];
+    let prefix = b"execute_proposal_with_signature_";
+    result[0..32].copy_from_slice(prefix);
+    result[32..64].copy_from_slice(&data_hash);
+    result
+}
```
shekohex, line 175: Same as above, we could better name this function:
```suggestion
pub fn make_ acknowledge_proposal_key(data_hash: [u8; 32]) -> [u8; 64] {
    let mut result = [0u8; 64];
    // make sure that this byte array is 32 bytes.
    let prefix = b"acknowledge_proposal_fixed_key_";
    result[0..32].copy_from_slice(prefix);
    result[32..64].copy_from_slice(&data_hash);
    result
}
```

event-watchers/substrate/src/signature_bridge_watcher.rs
```diff
@@ -389,11 +390,3 @@ fn secp256k1_ecdsa_recover(
...
     Ok(res)
 }
-
-fn make_execute_proposal_key(data_hash: [u8; 32]) -> [u8; 64] {
-    let mut result = [0u8; 64];
-    let prefix = b"execute_proposal_with_signature_";
-    result[0..32].copy_from_slice(prefix);
-    result[32..64].copy_from_slice(&data_hash);
-    result
-}
```
shekohex, line 399: we should revert this change and bring this function again back.

shekohex (changes requested):
Looks good! Almost complete, but not entirely right, yet! Left some comments.

event-watchers/substrate/src/signature_bridge_watcher.rs
Nutomic: I moved it to crates/proposal-signing-backends/src/dkg.rs above because I thought the code can be shared, but apparently thats not the case.

Nutomic:
Updated with these suggestions.

shekohex (approved):
LGTM! 👍

I will be auto-merged once all the CI checks passes.

Nutomic:
This test below keeps failing randomly, have to restart to fix it.

```
   1) Vanchor Transaction relayer
       number of deposits made should be equal to number of leaves in cache:

      AssertionError: expected [ …(10) ] to deeply equal [ …(10) ]
      + expected - actual
```

shekohex:
> This test below keeps failing randomly, have to restart to fix it.
> 
> ```
>    1) Vanchor Transaction relayer
>        number of deposits made should be equal to number of leaves in cache:
> 
>       AssertionError: expected [ …(10) ] to deeply equal [ …(10) ]
>       + expected - actual
> ```

Would you like to open a Task to look into this issue?

PR #472, opened 2023-04-20
Rework resource metrics (#472

### LINKED ISSUE #387: [SPEC] Improve Relayer metrics for better user readability
### Overview

We want to make use of the Relayer metrics for various things, for making it easier for DevOps and relayer operators to monitor and create custom alerts on top of the metrics we export. And the motivation here to make this possible by making it more readable for the end user.

As of now, we currently create/export metrics for (Total gas fee spent, Total fees earned, and account balance) per resource id, this all good except that this not really readable, the result metrics looks like this `resource_ 0000000000009678647b9fcb0039652a16dba688bd067d6e5077010000aa36a7_total_fees_earned` which is not really useful on its own. We need to think of a better way.

### Proposed solutions

TBD

### LINKED ISSUE #476: Metric `resource_account_balance` is never written to
This metric is only read, but never written. Thats why its always stuck at 0.

Nutomic:
## Summary of changes
- Previous Grafana dashboard was broken for some reason, had to recreate it from scratch
- Dashboard now includes resource metrics
- chain_account_balance was never written, added that now

- Closes #387
- Closes #476

-----
### Code Checklist 

- [ ] Tested
- [ ] Documented


Cargo.toml
```diff
@@ -43,7 +43,7 @@ sp-core = { version = "16.0.0", default-features = false }
 sp-runtime = { version = "18.0.0", default-features = false }
 # Used by ethers (but we need it to be vendored with the lib).
 native-tls = { version = "^0.2", features = ["vendored"] }
-webb-proposals = { version = "0.5.4", default-features = false, features = ["scale"] }
+webb-proposals = { git = "https://github.com/webb-tools/webb-rs", branch = "felix/display-resource-id", default-features = false, features = ["scale"] }
```
Nutomic, line 46: This is to include the Display trait for ResourceId from https://github.com/webb-tools/webb-rs/pull/133. However it fails to compile, dont know why:

```
error: cannot find macro `thread_local` in this scope
--> /home/felix/.cargo/git/checkouts/substrate-7e08433d4c370a21/1837f42/primitives/externalities/src/scope_limited.rs:22:1
|
22 | environmental::environmental!(ext: trait Externalities);
| ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
|
= note: `thread_local` is in scope, but it is an attribute: `#[thread_local]`
= note: this error originates in the macro `$crate::thread_local_impl` which comes from the expansion of the macro `environmental::environmental` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0425]: cannot find value `GLOBAL` in this scope
--> /home/felix/.cargo/git/checkouts/substrate-7e08433d4c370a21/1837f42/primitives/externalities/src/scope_limited.rs:22:1
|
22 | environmental::environmental!(ext: trait Externalities);
| ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not found in this scope
|
= note: this error originates in the macro `$crate::environmental` which comes from the expansion of the macro `environmental::environmental` (in Nightlybuilds, run with -Z macro-backtrace for more info)
```

We could leave this out for now and use the hex representation of resource for metrics until it is fixed.

crates/relayer-utils/src/metric.rs
```diff
@@ -169,34 +167,36 @@ impl Metrics {
...
-        let total_gas_spent_counter = register_counter!(
-            format!("resource_{resource_hex}_total_gas_spent"),
-            format!(
-                "The total number of gas spent on resource : {resource_hex}"
-            )
-        );
+        let total_gas_spent = register_counter!(opts!(
+            "resource_total_gas_spent",
+            "Total number of gas spent on resource",
+            labels!("resource_id" => &resource_name)
```
Nutomic, line 176: https://prometheus.io/docs/practices/naming/#labels

shekohex: I like that idea of using labels here for resource_ids!
what if we could even go further by making it more readable by having the following:
1. the API stays the same, it operates over the resource Ids, however, we could have many labels like
`chain_type` (`evm`, `substrate`, ..etc) and `chain_id` which is another label that represent the chainID.
`target_system_type` (`contract`, `tree_id`, ..etc) and `target_system_value` (`0xAe91...e28D`, 4, ..etc)

What do you think? we can now filter by labels like `evm` and chain id `5` with contract and its value `0xAe91...e28D`

crates/relayer-utils/src/metric.rs
```diff
@@ -169,34 +167,36 @@ impl Metrics {
...
-        let account_balance_counter = register_gauge!(
-            format!("resource_{resource_hex}_account_balance"),
-            format!("Total account balance : {resource_hex}")
-        );
+        let account_balance = register_gauge!(opts!(
+            "resource_account_balance",
+            "Total account balance",
+            labels!("resource_id" => &resource_name)
+        ))
+        .expect("create gauge for account balance");
```
shekohex, line 194: Here is the thing, the account balance is to tied to a single resoruce id, it is actually tied to a system or a chain, so we can move this up on its own `HashMap<TypedChainId, AccountBalanceMetric>` for example.

shekohex (changes requested):
Looks good! left some comments.

crates/relayer-utils/src/metric.rs
Nutomic: I will give this a try. Not entirely sure how it will work with Grafana if there are for example identical chain_ids on two different chains. But it can probably work somehow.

crates/tx-relay/src/substrate/mod.rs
```diff
@@ -114,3 +114,7 @@ pub async fn handle_substrate_tx(
     }
     Ok(())
 }
+
+fn wei_to_gwei(wei: u128) -> f64 {
+    (wei / (10 ^ 9)) as f64
+}
```
Nutomic, line 120: I didnt find any builtin function in subxt for this.

shekohex: We can use a similar approach with the one we did here #471, I guess it is okay to use ethers utils here.

Nutomic:
This is finished. Tests keep timing out in CI but they are passing locally.

PR #475, opened 2023-04-20
fix: DKG Signing Backend

shekohex:
## Summary of changes
- Use the chainId value from the pallet as the one for queueing transactions.
- Fixes the bug introduced by #461 

## Dynamic Transaction

There is an issue with creating dynamic transactions and `TypedChainId`, for some reason it does not work and returns this error:

```
Error encoding from dynamic value: Value shape is wrong; expected type ID 49, but got value Value { value: Composite(Named([("name", Value { value: Primitive(String("EVM")), context: () }), ("values", Value { value: Composite(Unnamed([Value { value: Primitive(U128(5001)), context: () }])), context: () })])), context: () }, which could not be coerced into it
```

**Logs**

```rust
2023-04-20T20:15:04.027954Z ERROR webb_probe: kind: tx_queue, ty: "SUBSTRATE", chain_id: 1081, errored: true, error: Error encoding from dynamic value: Value shape is wrong; expected type ID 49, but got value Value { value: Composite(Named([("name", Value { value: Primitive(String("EVM")), context: () }), ("values", Value { value: Composite(Unnamed([Value { value: Primitive(U128(5001)), context: () }])), context: () })])), context: () }, which could not be coerced into it
    at crates/tx-queue/src/substrate/substrate_tx_queue.rs:140
    in webb_relayer_tx_queue::substrate::substrate_tx_queue::run with node: 1081
```

**What I have tried and did not work**

- [x] `Value::from_bytes(src_chain_id.encode())`
- [x] `Value::u128(u128::from(self.src_chain_id.chain_id())`

- Closes #474 

-----
### Code Checklist 

- [ ] Tested
- [ ] Documented

### DIFF
--- a/crates/proposal-signing-backends/src/dkg.rs
+++ b/crates/proposal-signing-backends/src/dkg.rs
@@ -14,34 +14,25 @@ use webb::substrate::subxt::dynamic::Value;
 use webb_relayer_store::{QueueStore, SledStore};
 use webb_relayer_store::sled::SledQueueKey;
 use webb_relayer_types::dynamic_payload::WebbDynamicTxPayload;
-use webb::substrate::subxt::tx::{PairSigner};
+use webb::substrate::subxt::tx::PairSigner;
 use webb::substrate::dkg_runtime::api::runtime_types::sp_core::bounded::bounded_vec::BoundedVec;
 
 type DkgConfig = PolkadotConfig;
 type DkgClient = OnlineClient<DkgConfig>;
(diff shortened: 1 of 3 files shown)


shekohex:
Related Repos and examples:
- https://github.com/paritytech/scale-value
- https://github.com/paritytech/scale-value/blob/f04c1afb244a7ff47208a0b82907a0b6cc9f7e85/src/scale_impls/encode.rs#L392-L402 (This how to encode an enum variant)

PR #483, opened 2023-04-27
Try using shared key for rust cache to speedup CI

Nutomic:
## Summary of changes
- Set `shared-key` for Rust Cache. This decreases Rust built time for integration tests from 8m to 50s. So the cache was really not used before.
- Tried caching `dvc` folders. This decreases duration of "Install NPM packages" (which includes `dvc pull`) from 4m 40s to 2m 50s. It might also help with occasional failures from dvc pull. However it increases cache size from 700 MB to 3.5 GB so the 10 GB limit can store only 3 cache entries instead of 14. Not sure if this is a good idea.
- Tried to add caching for unit test and check commands, but this fails due to usage of `cargo cross`:
    ```
    error: failed to run custom build command for `libc v0.2.141`
    
    Caused by:
      process didn't exit successfully: `/target/debug/build/libc-419af779436d366a/build-script-build` (exit status: 1)
      --- stderr
    /target/debug/build/libc-419af779436d366a/build-script-build: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.29' not found (required by /target/debug/build/libc-419af779436d366a/build-script-build)
    /target/debug/build/libc-419af779436d366a/build-script-build: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by /target/debug/build/libc-419af779436d366a/build-script-build)
    warning: build failed, waiting for other jobs to finish...
    Error: Process completed with exit code 101.
    ```
    [Upstream issue is here](https://github.com/Swatinem/rust-cache/issues/42). Anyway these jobs are fast enough with around 10m so its probably fine without caching.  

- Closes 

-----
### Code Checklist 

- [ ] Tested
- [ ] Documented

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -114,6 +114,11 @@ jobs:
           toolchain: stable
           targets: ${{ matrix.target }}
 
+      - name: Rust Cache
+        uses: Swatinem/rust-cache@v2
+        with:
+          shared-key: "rust"
+
       - name: Install cross
         run: cargo install cross --locked
       - name: Check target using cross
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -42,7 +42,7 @@ jobs:
         uses: Swatinem/rust-cache@v2
         with:
           shared-key: "rust"
-          cache-directories: ".dvc/tmp\n.dvc/"
+          cache-directories: ".dvc/tmp\n.dvc/cache"
 
       - name: Build target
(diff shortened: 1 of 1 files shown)


shekohex:
Awesome work! Really looking Forward to this to be merged!

Nutomic:
Should we also apply these changes to our other Rust repos?

shekohex:
> Should we also apply these changes to our other Rust repos?

Lets keep it for the relayer repo for now, if we had a success with it we can let other team use this technique too.

PR #497, opened 2023-05-09
bugfix: Custom Retry Policy for EVM providers

### LINKED ISSUE #496: [TASK] Improve Network Connection Reliability for EVM Chains

shekohex:
## Summary of changes
- Add Custom Retry Policy for EVM Providers
- Add Tests to ensure the custom retry policy works as intended.

- Closes #496 

-----
### Code Checklist 

- [x] Tested
- [x] Documented

### DIFF
--- a/crates/relayer-context/src/lib.rs
+++ b/crates/relayer-context/src/lib.rs
@@ -43,6 +43,9 @@ use webb_price_oracle_backends::{
 use webb_relayer_store::SledStore;
 use webb_relayer_utils::metric::{self, Metrics};
 
+mod ethers_retry_policy;
+use ethers_retry_policy::WebbHttpRetryPolicy;
+
 /// RelayerContext contains Relayer's configuration and shutdown signal.
 #[derive(Clone)]
 pub struct RelayerContext {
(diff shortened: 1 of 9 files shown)


crates/relayer-context/src/lib.rs
```diff
@@ -143,8 +146,10 @@ impl RelayerContext {
         })?;
         let client = Http::new(chain_config.http_endpoint.clone());
         // Wrap the provider with a retry client.
-        let policy = Box::new(providers::HttpRateLimitRetryPolicy);
-        let retry_client = RetryClientBuilder::default().build(client, policy);
+        let retry_client = RetryClientBuilder::default()
+            .timeout_retries(u32::MAX)
+            .rate_limit_retries(u32::MAX)
```
Nutomic, line 151: This means it makes unlimited retries right? Is that really a good idea? It could take up a lot of resources over time if there are hundreds or thousands of requests to retry. I would retry like 3-5 times max and then log a warning if it still fails.

Nutomic (approved):

crates/relayer-context/src/lib.rs
shekohex: I do not like it either, but that's all we have since the response does not contain how much secs we should backoff and retry again. Right now we default to 5s, and we hope the service/rate limit is lifted and we can send more requests.

shekohex: in the future, we will be running our own nodes or we use a premium RPC endpoints.

PR #501, opened 2023-05-11
feature: Optmize the usage of RPCs and Resoruces

shekohex:
## Highlights of this PR

- [Use of the cached `chain_id` when possible instead of refetching it](https://github.com/webb-tools/relayer/commit/3622570f066696682175d7d4385659fbedceaa5a)
previously, we used to call `eth_chainId` on every block/event iterations, we do not need this at all, we can cache it or pass it to the service that needs it.
- [Make the default `max_blocks_per_step` is `500` blocks/step](https://github.com/webb-tools/relayer/commit/75e821ff853bd73d320248fda701ea0fab7b2ffb)
This should make the initial sync faster by default (you still can override it from the config)
- [Correctly build the merkle tree in the leaves handler](https://github.com/webb-tools/relayer/commit/6ee7412c53ca1a6691ea1ae774363d371ee6ee5c)
There was a bug where when we start back up, we do not initialize the merkle tree with the saved leaves.
- [Cache the EVM Providers instead of creating ones every single time](https://github.com/webb-tools/relayer/commit/17dc90c56c1ae8ba122b91abe505f2370eab37bc)
Everytime we call `evm_provider` we were creating a new connection to the RPC node, now we can cache these instead.
- [Change the way we return the cached leaves](https://github.com/webb-tools/relayer/commit/3331605f628142277a040747c7dce6bd7744dce6)
It is now retruning as a `leaf_index -> leaf_bytes`. However, this does not affect the public leaves API (so no breaking changes).
- [Do not fetch the chainId nor the latest block number when not needed](https://github.com/webb-tools/relayer/commit/e0b0340f1da39537cd6df635919f4a95a752d0c1)
This a minor optimization, where we first check if we can handle this event or not, before featching the latest block number to check if we are syncing or not. There is a room to even optmize this more.
- [Optmize the way we fetch events.](https://github.com/webb-tools/relayer/commit/b4e84bdf611f1be82237956b1d1c9520b36ea448) 
**This mostly the biggest changes here, So I'm going to explain it as detailed as possible**:

**TL;DR version**: The new implementation offers better performance and reliability by reducing the load on system resources and ensuring synchronization with the latest block.

**The Long version**: The previous implementation was inefficient as it needlessly consumed a lot of resources. It performed a fetch of the last block number at each iteration, calculated the destination block number, and then fetched the events. The process continued in a loop without a cooldown until the latest block was reached. The system heavily relied on this process, which caused a significant drain on resources.

To remedy this issue, the new implementation fetches the last block number once and never updates it until it is fully synchronized with the latest block. This new implementation allows us to fetch the events in fewer calls without constant polling until the latest block is reached. Only then can we confirm that the system is fully synchronized. Until this point, the system only updates the latest block, making it more efficient in its resource utilization.

It is important to note that with the new implementation, fetching the latest block is no longer in a hot loop. This is because the system only fetches it after a `polling interval` of milliseconds. This makes the system more efficient and reduces the load on the resources.

Overall, the new implementation is significantly more efficient and reliable than the previous one. It allows for a better use of system resources and ensures that the system stays synchronized with the latest block.

-----
### Code Checklist 

- [x] Tested
- [x] Documented

### DIFF
--- a/.github/workflows/image-publish.yml
+++ b/.github/workflows/image-publish.yml
@@ -58,16 +58,12 @@ jobs:
       - name: install system build dependencies
         run: sudo apt-get update && sudo apt-get install ${DEV_PACKAGES}
       - uses: actions/checkout@v2
-      - name: Cache Cargo
-        uses: actions/cache@v2
+      - name: Rust Cache
+        uses: Swatinem/rust-cache@v2
         with:
-          path: |
-            ~/.cargo/registry
-            ~/.cargo/git
-            ~/.cargo/bin
-            target/x86_64-unknown-linux-musl
-            target/release
-          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}-x86_64-unknown-linux-musl
+          shared-key: "rust"
+          cache-directories: ".dvc/tmp\n.dvc/cache"
+
       - uses: actions/download-artifact@v1
         with:
           name: cross-linux-musl
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -53,15 +53,11 @@ jobs:
         with:
           access_token: ${{ github.token }}
       - uses: actions/checkout@v2
-      - name: Cache Cargo
-        uses: actions/cache@v2
+      - name: Rust Cache
+        uses: Swatinem/rust-cache@v2
         with:
-          path: |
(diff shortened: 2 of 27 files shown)


salman01zp (approved):
LGMT..!

Issue #482, opened 2023-04-27
[TASK] Create a new release from the `develop` branch


shekohex:
### Overview

Similar to [rust-analyzer/releases/nightly](https://github.com/rust-lang/rust-analyzer/releases/tag/nightly) that gets updated every night, for us we should create a new `develop` release every time we merge a PR to the `develop` branch.

### Checklist

- [ ] Adopt [github-release](https://github.com/rust-lang/rust-analyzer/tree/master/.github/actions/github-release) action.
- [ ] Customize it for our needs (should not be a lot of changes.)
- [ ] Create the `develop` release on every new PR gets merged to the `develop` branch.

PR #502, opened 2023-05-12
Automatic release when pushing to develop branch

Nutomic:
## Summary of changes
- 

- Closes https://github.com/webb-tools/relayer/issues/482

-----
### Code Checklist 

- [ ] Tested
- [ ] Documented

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -125,7 +125,7 @@ jobs:
         with:
           path: dist
       #- run: ls -al ./dist/**
-      - run: ls -al .github/actions/github-release
+      - run: ls -al .github
       - run: echo ${{github.action}}
       - name: Publish Release
         uses: ./.github/actions/github-release
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -111,7 +111,7 @@ jobs:
   publish:
     name: publish
     runs-on: ubuntu-latest
-    needs: ["macos", "linux"]
+    #needs: ["macos", "linux"]
     env:
       # TODO: change to develop branch
       release_name: ${{ (github.branch == 'felix/automatic-release') && 'develop' || github.ref}}
@@ -124,7 +124,8 @@ jobs:
         uses: actions/download-artifact@v3
         with:
           path: dist
-      - run: ls -al ./dist
+      - run: ls -al ./dist/**
+      - run: ls -al .github/actions/github-release
       - run: echo ${{github.action}}
       - name: Publish Release
         uses: ./.github/actions/github-release
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -111,7 +111,7 @@ jobs:
   publish:
     name: publish
     runs-on: ubuntu-latest
-    #needs: ["macos", "linux"]
+    needs: ["macos", "linux"]
     env:
       # TODO: change to develop branch
       release_name: ${{ (github.branch == 'felix/automatic-release') && 'develop' || github.ref}}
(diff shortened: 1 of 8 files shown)


.github/workflows/release.yml
```diff
@@ -95,6 +90,8 @@ jobs:
           - aarch64-unknown-linux-gnu
           - x86_64-unknown-linux-gnu
           - x86_64-unknown-linux-musl
+          - aarch64-unknown-linux-
```
shekohex, line 93:
```suggestion
          - aarch64-unknown-linux-musl
```

.github/workflows/release.yml
```diff
@@ -117,22 +114,42 @@ jobs:
...
-        uses: actions/upload-release-asset@v1
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          name: dist-aarch64-pc-windows-msvc
+          path: |
+            webb-relayer-aarch64-apple-darwin.tar.gz
+            webb-relayer-aarch64-unknown-linux-gnu.tar.gz
+            webb-relayer-x86_64-apple-darwin.tar.gz
+            webb-relayer-x86_64-unknown-linux-gnu.tar.gz
+            webb-relayer-x86_64-unknown-linux-musl.tar.gz
```
shekohex, line 143: Maybe use the same `${{ env.BIN }}-${{ matrix.target }}.tar.gz` syntax here?

.github/workflows/release.yml
```diff
@@ -117,22 +114,42 @@ jobs:
...
-          tag_name: ${{ github.ref }}
-          release_name: ${{ github.ref }}
-          draft: false
-          prerelease: false
-      - name: Upload Release Asset
-        id: upload-release-asset
-        uses: actions/upload-release-asset@v1
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          name: dist-aarch64-pc-windows-msvc
```
shekohex, line 137: What is this? we do not support windows targets for now.

shekohex (changes requested):

.github/workflows/release.yml
Nutomic: These matrices are defined inside macos/linux tasks, so I dont think they are available from here.

.github/workflows/release.yml
Nutomic: Leftover from copy paste. Anyway this is probably all broken, will need to test it.

.github/workflows/release.yml
```diff
@@ -1,4 +1,11 @@
...
+  # separate push conditions so they are combined with 'or', not 'and'
+  # TODO: this is not working, might have to run on all pushes and then cancel manually?
+  push:
+    branches:
+      - develop
+      # TODO: only for testing
+      - felix/automatic-release
   push:
     # Sequence of patterns matched against refs/tags
     tags:
```
Nutomic, line 11: It seems that multiple `push` values are always combined with AND, so the workflow is only excecuted if the push is on branch develop AND it is a tag. So far I havent found a way to make an OR condition, other than executing unconditionally and then checking manually if it should continue or abort.

shekohex: There is also this option, where we it runs if a PR got merged: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-pull_request_target-workflow-when-a-pull-request-merges

Since we can't just push to the `develop` branch without a PR, this could also be an option.

Nutomic: I think that would still run on the PR branch so its probably not what we want. I will try running the workflow always on main and develop branches, then checking manually for tags on main.

shekohex (commented):
Could we change the name `nightly` every where to be `develop`? we are not going to makea release really every "night". it is just that every time we merge a PR into the `develop` branch.

PR #517, opened 2023-05-17
Fix partial writes

### LINKED ISSUE #506: [TASK] Use Transactional and Atomic database operations when possible
### Overview

Right now, we rarely make use of the atomic updates when we are working with the database, this not something we should relay on, and instead we should use the transactional API when possible.

### The Problem

Partial writes is the root issue here, we should identify and group related operations into one atomic transaction to the database, for example, when we save the leaf and its index, then we save the last block number of that event. It happens that the relayer/user stops at the moment just before we save the last block number. This means we are going to fetch the same event again, which is not the main issue, but could lead to the bugs where we insert the same leaf again into the local merkle tree, making it producing an invalid root. Solution would be to make these two ops into one atomic transaction.

### Task Checklist

This checklist are split into two parts:

- **The first part**:
- [ ] Research and identify all the operations that we are using the database for and list them below.
- [ ] Find and identify the operations that should be grouped together into one atomic operation.

- **The second part**
- [ ] Make the required changes to the database interface to allow the above changes.
- [ ] Add more tests to make sure this works.
- [ ] Update the docs

salman01zp:
## Summary of changes
- Fix partial write issue with computing Merkle root in `can_handle_evensts`, moved it to `handle_events`
- Group operation for LeafStoreCache traits using sled atomic transactions
https://docs.rs/sled/latest/sled/transaction/index.html

```rust
store.insert_leaves()
// This is the last deposit event block number
store.insert_last_deposit_block_number()
// This will be used by the event watcher to track the block number that has been processed
store.set_last_block_number()
```

- Closes #506 

-----
### Code Checklist 

- [x] Tested
- [ ] Documented

### DIFF
--- a/crates/relayer-store/src/lib.rs
+++ b/crates/relayer-store/src/lib.rs
@@ -285,13 +285,6 @@ pub trait LeafCacheStore: HistoryStore {
         range: core::ops::Range<u32>,
     ) -> crate::Result<Self::Output>;
 
-    /// Insert the leaves for the given key.
-    fn insert_leaves<K: Into<HistoryStoreKey> + Debug>(
-        &self,
-        key: K,
-        leaves: &[(u32, Vec<u8>)],
-    ) -> crate::Result<()>;
-
     /// The last deposit info is sent to the client on leaf request
     /// So they can verify when the last transaction was sent to maintain
     /// their own state of mixers.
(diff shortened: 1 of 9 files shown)


shekohex (changes requested):
Looks good! 

I left a few comments, one of them actually I did forgot to add which was Do you think we should keep the old `store_*` methods? if not, we shall remove them and keep the atomic ones instead, to not forget and use the old ones.

shekohex (approved):
LGTM!

PR #519, opened 2023-05-22
Update relayer endpoints

### LINKED ISSUE #508: [TASK] Return Cached Leaves as different format instead of Arrays
### Overview

Currently the relayer returns the cached leaves/encrypted outputs as a bytes array which is not optimal size-wise, for comparison:

```js
// This just one leaf and the total size is 118 bytes
[0,96,124,186,197,202,118,215,202,188,115,135,221,48,154,82,56,200,222,123,134,81,222,24,29,150,89,62,147,193,111,174]
// the same exact leaf and the total size is 66 bytes.
"00607cbac5ca76d7cabc7387dd309a5238c8de7b8651de181d96593e93c16fae"
// Even to reduce the size further, we could reduce the size even more by using base64 (instead of hex base16)
// to just 46 bytes
"AGB8usXKdtfKvHOH3TCaUjjI3nuGUd4YHZZZPpPBb64="
```

- Using Arrays (Baseline) = 118 bytes.
- Using Hex strings = 66 bytes (~178% Smaller)
- Using base64 strings = 46 (~256% Smaller)

### Task Checklist

- [ ] Decide which format we are going to utilize.
- [ ] Update the leaves endpoint (For EVM/Substrate)
- [ ] Update the Encrypted outputs endpoint (For EVM/Substrate)
- [ ] Update the tests
- [ ] Update the docs (README, docs site, ..etc)
- [ ] Notify the dApp team with the breaking changes in the PR.

### LINKED ISSUE #510: [TASK] Remove the need for `TypedChainId`s in the API Endpoints
### Overview

We should be consistent while working and designing our APIs, right now I see a divergence when we query something, for instance when we query the leaves cache or the encrypted outputs, we use the raw chain ids like 5 for Goerli, but when we work with fee info we use the typed chain Ids, which is not human-friendly esp for testing and debugging.

Since now we have two separate Endpoints for Fee informations, we should change the endpoints to use the raw untyped chain ids like the other endpoints we have. 

https://github.com/webb-tools/relayer/blob/008b8225a88bcb5ae4a35cb28697fbf4b9c92c02/crates/relayer-handlers/src/lib.rs#L200-L237

### Task Checklist

- [x] Change EVM Fee info to accept the `chain_id` instead of `typed_chain_id`
- [x] Change Substrate Fee info to accept the `chain_id` instead of `typed_chain_id`
- [x] Update tests
- [x] Update docs (README, docs website, ..etc)
- [ ] Notify the dApp team with these breaking changes before we deploy the new update.

### LINKED ISSUE #512: [TASK] Embed Relayer version, git commit/rev into relayer binary for `/api/v1/info`
### Overview

It is hard to tell what version or what git commit this relayer is currently running? which makes me think we should embed these information in the result binary at compile-time so that we can return these info in the `/api/v1/info` endpoint. This will be super useful in the future (for now here is a tease for future usage of this: `relayerup clone https://relayer1.webb.tools` :shushing_face: )

### Example

By default, most of the information we add in Cargo.toml are already exposed as `CARGO_*` env during compile time, for example:

```rust
let version = env!("CARGO_PKG_VERSION");
```
For more, you can read about it [here](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates).

- **What about git commit and git release name?**
In our build.rs we can do something like the following:

```rust
println!("cargo:rustc-env=GIT_BRANCH={}", get_git_branch()));
```

This will tell cargo to set this env for us, so that we can do the following in our main program:

```rust
let branch = env!("GIT_BRANCH");
```

> There a crate that actually helps in doing all of this [`build-data`](https://docs.rs/crate/build-data/0.1.4)

### Expected Response

Utilizing this in Our CI/CD, and reading that in our code, we want to have something like the following in our `/api/v1/info`

```json
{
  "evm": {},
  "substrate": {},
  "experimental": {},
  "features": {},
  "assets": {},
  "build": {
    "version": "0.5.0-rc22-alpha",
    "commit": "ef4e899",
    "timestamp": "2023-05-12T13:42:54Z"
  }
}
```
> NOTE: the `build.timestamp` uses `SOURCE_TIMESTAMP` env, to help making this a [`Reproducible build`](https://reproducible-builds.org/docs/timestamps/)

### Task Checklist

- [ ] Make use of `build-data` crate
- [ ] Embed the required information in the `build.rs`
- [ ] Add a CI check in the release workflow that checks that version inside the `Cargo.toml` matches the version of the tag relesae.
- [ ] Update the `/api/v1/info` endpoint.
- [ ] Update the docs and tests if required.

salman01zp:
## Summary of changes
- **Leaf Caching API**
    Leaf caching API will be sending cached leaf as `HexString` instead of bytes.
    ```
       /api/v1/leaves/evm/:chain_id/:contract_address
    ```
     Response
     ```json
          {
              "leaves": [
                  "0x2f286532136e3110a63779df0cac44543d10d94e54f651e86eaff7851a631e45",
                  "0x20f9563640ed1d21151a60823212499497ef6b3c0d2a6593d6a2a9748fc1db98",
                  "0x283376a2134ac8096e523c74573dc84eec9c308ee9f831595b25b1ecb01f78a0",
                  "0x1c22842f1b1b731d5a3d9f3b0fe7de1ba7f1ede9eadc3b699121363edd337108",
                  "0x0aebac8baafcba5873723eb58b947c9c5f523a28c1c5872a613febc9942b0821",
                  "0x2268d2119c2d5de58cd5367a54549b07decc9d299f1e9b1810c0926977c915f4",
                  "0x194e2540861267b9a3d31660e242fc40774c9a72c32b7f67a7fcd4dc9193102e",
                  "0x0784707003bd770b2a57f02434c4fa5035f5777f57750d6c91fccf384418ef5e",
                  "0x1962dda78713ae31c2d5b7ec727c10ff61acc918e7c76e570bc25d9b73e58f22",
                  "0x0d9ad8106a87ea871b8bc1620388fa35a95a0201f9b94746e87d10f9b0f53d6b"
                ],
  
         }
     ```
---
- **Fee Info API**
   Since we made separate  API into evm and substrate, we don't need to send TypedChainIds anymore but instead, need to send underlying chain id. eg 5001, 5002
   EVM
   ```
       /api/v1/fee_info/evm/:chain_id/:vanchor/:gas_amount
   ```
   Substrate
   ```
      /api/v1/fee_info/substrate/:chain_id/:partialFee
   ```
---

-  **Relayer Info API**
    Adding build info configuration here, so we can get build-related details like build version, branch, commit hash, build time.
     ```
       /api/v1/info
     ```
     Response
     ```json
         {
            "evm": {
                "rinkeby": {
                    "enabled": true,
                    "chainId": 4,
                    "beneficiary": "0x58fcd47ece3ed24ace88fee06efd90dcb38f541f",
                    "contracts": [{
                        "contract": "Anchor",
                        "address": "0x9d36b94f245857ec7280415140800dde7642addb",
                        "deployedAt": 8896800,
                        "eventsWatcher": {
                            "enabled": true,
                            "pollingInterval": 15000
                        },
                        "size": 0.1,
                        "proposalSigningBackend": { "type": "DKGNode", "node": "dkg-local" },
                        "withdrawFeePercentage": 0.05
                    }]
                }
            },
            "substrate": {},
            "experimental": {
                "smart-anchor-updates": false,
                "smart-anchor-updates-retries": 0
            },
            "features": { "dataQuery": true, "governanceRelay": true, "privateTxRelay": true },
            "assets": {
                "TNT": { "price": 0.1, "name": "Tangle Network Token", "decimals": 18 },
                "tTNT": { "price": 0.1, "name": "Test Tangle Network Token", "decimals": 18 }
            },
            "build" : {
                "version": "0.5.0",
                "commit": "c8875ba78298d34272e40c2e302fcfe33f191147",
                "build_time": "2023-05-19T15:57:40Z"
            }
        }
    ```
   
## Note For Dapp Team
Since this will be breaking changes, you will need to make appropriate changes in DAPP.
 CC: @AtelyPham 

- Closes #508 
- Closes #510 
- Closes #512 

-----
### Code Checklist 

- [x] Tested
- [x] Documented

### DIFF
--- a/README.md
+++ b/README.md
@@ -294,7 +294,7 @@ The relayer has 3 endpoints available to query from. They are outlined below for
     "build" : {
         "version": "0.5.0",
         "commit": "c8875ba78298d34272e40c2e302fcfe33f191147",
-        "build_time": "2023-05-19T15:57:40Z"
+        "timestamp": "2023-05-19T15:57:40Z"
     }
 }
   ```
--- a/README.md
+++ b/README.md
@@ -261,8 +261,7 @@ The relayer has 3 endpoints available to query from. They are outlined below for
   <summary>Expected Response</summary>
   
   ```json
-{
-  "relayer_config" : {
+  {
     "evm": {
         "rinkeby": {
             "enabled": true,
@@ -286,13 +285,16 @@ The relayer has 3 endpoints available to query from. They are outlined below for
     "experimental": {
         "smart-anchor-updates": false,
         "smart-anchor-updates-retries": 0
-    }
-  },
-  "build_info" : {
-    "version": "0.5.0",
-    "commit": "c8875ba78298d34272e40c2e302fcfe33f191147",
-    "branch": "develop",
-    "build_time": "2023-05-19T15:57:40Z"
+    },
+    "features": { "dataQuery": true, "governanceRelay": true, "privateTxRelay": true },
+    "assets": {
+        "TNT": { "price": 0.1, "name": "Tangle Network Token", "decimals": 18 },
+        "tTNT": { "price": 0.1, "name": "Test Tangle Network Token", "decimals": 18 }
+    },
+    "build" : {
+        "version": "0.5.0",
+        "commit": "c8875ba78298d34272e40c2e302fcfe33f191147",
+        "build_time": "2023-05-19T15:57:40Z"
     }
 }
   ```
--- a/README.md
+++ b/README.md
@@ -261,7 +261,8 @@ The relayer has 3 endpoints available to query from. They are outlined below for
   <summary>Expected Response</summary>
   
   ```json
-  {
+{
+  "relayer_config" : {
     "evm": {
         "rinkeby": {
             "enabled": true,
@@ -285,6 +286,13 @@ The relayer has 3 endpoints available to query from. They are outlined below for
     "experimental": {
         "smart-anchor-updates": false,
         "smart-anchor-updates-retries": 0
+    }
+  },
+  "build_info" : {
+    "version": "0.5.0",
+    "commit": "c8875ba78298d34272e40c2e302fcfe33f191147",
+    "branch": "develop",
+    "build_time": "2023-05-19T15:57:40Z"
     }
 }
   ```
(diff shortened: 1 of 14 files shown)


crates/relayer-handlers/src/routes/info.rs
```diff
@@ -10,12 +10,35 @@ use webb::evm::ethers::{
...
 
+/// Build info data
+#[derive(Debug, Serialize)]
+pub struct BuildInfo {
+    /// Version of the relayer
+    pub version: String,
+    /// Commit hash of the relayer
+    pub commit: String,
+    /// Branch of the relayer
+    pub branch: String,
```
shekohex, line 21: I do not think the branch name would make sense in the result binary, if we are building of a git tag branch, it will just say "HEAD", so we will probably remove it, a git commit and a version is all we need.

shekohex (changes requested):
Looks good! I left a few comments. Please address these changes and update the docs.

shekohex:
Also lets from now on update our version in the following https://semver.org/
https://github.com/webb-tools/relayer/blob/c8875ba78298d34272e40c2e302fcfe33f191147/services/webb-relayer/Cargo.toml#L3

Examples:
* For develop branch, we must always use the term `dev` at the end of the version.
```
0.5.19-dev
```
* when creating a new pre-release, we remove the `dev` part and start using `rc.X` part where we increment the `X` each time:
```
0.5.19-rc.25
```
* When we create a stable release, or while we are going to push to a `main` branch, we take the last `rc.X` and use it as stable:
```
0.5.19
```

crates/relayer-handlers/src/routes/info.rs
```diff
@@ -10,12 +10,34 @@ use webb::evm::ethers::{
...
 
+/// Build info data
+#[derive(Debug, Serialize)]
+pub struct BuildInfo {
+    /// Version of the relayer
+    pub version: String,
+    /// Commit hash of the relayer
+    pub commit: String,
+    /// Build time of the relayer
+    pub build_timestamp: String,
```
shekohex, line 21:
```suggestion
    pub timestamp: String,
```
It reads better in JSON.

README.md
```diff
@@ -285,6 +285,16 @@ The relayer has 3 endpoints available to query from. They are outlined below for
...
+    },
+    "features": { "dataQuery": true, "governanceRelay": true, "privateTxRelay": true },
+    "assets": {
+        "TNT": { "price": 0.1, "name": "Tangle Network Token", "decimals": 18 },
+        "tTNT": { "price": 0.1, "name": "Test Tangle Network Token", "decimals": 18 }
+    },
+    "build" : {
+        "version": "0.5.0",
+        "commit": "c8875ba78298d34272e40c2e302fcfe33f191147",
+        "build_time": "2023-05-19T15:57:40Z"
```
shekohex, line 297:
```suggestion
        "timestamp": "2023-05-19T15:57:40Z"
```

shekohex (commented):
One more thing and we good to go.

shekohex (approved):
LGTM!

PR #525, opened 2023-05-29
fix: Move tx signing logic to the substrate transaction queue

### LINKED ISSUE #524: [BUG] Substrate Transaction Queue fails silently

shekohex:
## Summary of changes
- Add Typed Erased Static Transaction payload so we can serialize it without limitations.
- Move the signing logic to the substrate tx queue, so that we do not need to sign anything outside the queue or manage the nonce.
- Make substrate chain id generic (to allow the method to accept any type that could be converted to a chain id)
- Improve the logging for better debugging. 
- Added a test that could be run locally to test the tx queue.

- Closes #524 

-----
### Code Checklist 

- [x] Tested
- [x] Documented


salman01zp (commented):
You will also need to enqueue the `TypeErasedStaticTxPayload` transaction for substrate SignatureBridgeWatcher

salman01zp (approved):
LGTM..!

shekohex:
@salman01zp any idea why this test keeps failing? It does pass locally.

shekohex:
Looks like the Node just hangs after 30 blocks, does not make any progress. Running the latest Tangle from the main branch.
image

PR #529, opened 2023-06-01
Multiple Rpc support for Evm chains

### LINKED ISSUE #499: [TASK] Add support for multi-provider for EVM networks
### Overview

Relayers right now for each chain they relay on a single RPC,  Relayers should make use of many of them so we improve the reliablity.

### Checklist

- [ ] Allow the configurations for RPC endpoints to be either one endpoint or a list of endpoints
- [ ] Relayers should make use of all endpoints by either testing them first and then in a round-robin fashion.
- [ ] Add tests (which includes bad endpoints for testing)
- [ ] Update docs
- [ ] Update config files

salman01zp:
### 1. Update EvmChainConfig to take multiple http-endpoints
```rust
pub struct EvmChainConfig {
    #[serde(skip_serializing)]
    pub http_endpoints: Vec<RpcUrl>,
}
```

### 2. Make evm_provider  mutable so we can update and cache latest working client

```rust
pub struct RelayerContext {
  .
  .
 /// Evm Providers Cache.
 evm_providers: Arc<Mutex<HashMap<types::U256, Arc<EthersClient>>>>,
}


pub async fn evm_provider<I: Into<types::U256>>(
        &self,
        chain_id: I,
    ) -> webb_relayer_utils::Result<Arc<EthersClient>> {
        let chain_id: types::U256 = chain_id.into();
        // Check if the provider is already in the cache or else create new provider from http_endpoint array.
        // If the provider is not working, try creating a new provider from http_endpoint array.

        if let Some(provider) = self.evm_providers.lock().await.get(&chain_id) {
            // get gas price to check if the provider is working
            let response = provider.get_gas_price().await;
            match response {
                Ok(_) => Ok(provider.clone()),
                Err(e) => {
                    tracing::error!(
                        "EVM Provider for chain {} is not working: {}.Try Connecting with other http_endpoints, if any.",
                        chain_id,
                        e
                    );
                    self.create_evm_provider(chain_id).await
                }
            }
        } else {
            self.create_evm_provider(chain_id).await
        }
    }


```

### 3. Update retry count for  RetryClient.

Since we want to switch between RPC providers, we should define a smaller limit for retires like ~50.  Once retry limit exceeds the current client is dropped and a new client is created by iterating over available rpc endpoints.
This will happen in the event_watcher::run() method

The `evm_provider` method will check the provider cache for the client and will do a sanity check for connection by making an API call. If the connection fails it will try to create a new evm_provider from the set of RPC endpoints in the loop.

```rust

let task = || async {
            let step = contract.max_blocks_per_step().as_u64();
            let provider = ctx.evm_provider(chain_id).await?;
            let client = Arc::new(TimeLag::new(
                provider,
                chain_config.block_confirmations,
            ));

     // drop client if retries exceed and create new clients after some back off time
}
```

- Closes #499 

-----
### Code Checklist 

- [x] Tested
- [ ] Documented

### DIFF
--- a/crates/relayer-context/src/lib.rs
+++ b/crates/relayer-context/src/lib.rs
@@ -159,7 +159,7 @@ impl RelayerContext {
                 Ok(_) => Ok(provider.clone()),
                 Err(e) => {
                     tracing::error!(
-                        "EVM Provider for chain {} is not working: {}.Connecting with other http_endpoints, if any.",
+                        "EVM Provider for chain {} is not working: {}.Try Connecting with other http_endpoints, if any.",
                         chain_id,
                         e
                     );
@@ -177,6 +177,8 @@ impl RelayerContext {
         &self,
         chain_id: I,
     ) -> webb_relayer_utils::Result<Arc<EthersClient>> {
+        use webb::evm::ethers::prelude::rand::Rng;
+
         let chain_id: types::U256 = chain_id.into();
         let chain_config =
             self.config.evm.get(&chain_id.to_string()).ok_or_else(|| {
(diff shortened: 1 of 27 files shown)


crates/relayer-context/src/lib.rs
```diff
@@ -70,7 +70,7 @@ pub struct RelayerContext {
     etherscan_clients: Arc<HashMap<types::U256, ethers::etherscan::Client>>,
 
     /// Evm Providers Cache.
-    evm_providers: Arc<HashMap<types::U256, Arc<EthersClient>>>,
+    evm_providers: Arc<Mutex<HashMap<types::U256, Arc<EthersClient>>>>,
```
shekohex, line 73: This does not have to be mutable? why it needs the lock?
for each network, we will maintain a list of providers and then switch between them in a round-robin fashion. Right now I do not see this happening anywhere.

crates/relayer-context/src/lib.rs
```diff
@@ -161,15 +149,86 @@ impl RelayerContext {
...
+                }
+            }
         } else {
-            let chain_id_string = chain_id.clone().to_string();
-            Err(webb_relayer_utils::Error::ChainNotFound {
-                chain_id: chain_id_string,
-            })
+            self.create_evm_provider(chain_id).await
+        }
+    }
```
shekohex, line 172: Again, we do not need this at all, we already have a RetryClient wrapper implemented. we need all of the providers to be added, no matter what if they are working or not.

shekohex (changes requested):
@salman01zp Please stick to the original [spec](https://github.com/webb-tools/relayer/issues/499), your current implementation does not adhere to any of the original goal of the specification we wrote. if anything is not clear in the original spec issue, please ask and I will try to clarify it for you.

salman01zp:
Will update in new PR

PR #531, opened 2023-06-05
Bundle Multiple EVM Providers 

salman01zp:
## Summary of changes
- 

- Closes 

-----
### Code Checklist 

- [ ] Tested
- [ ] Documented

### DIFF
--- a/crates/relayer-context/src/lib.rs
+++ b/crates/relayer-context/src/lib.rs
@@ -45,7 +45,7 @@ use webb_relayer_utils::metric::{self, Metrics};
 mod ethers_retry_policy;
 use ethers_retry_policy::WebbHttpRetryPolicy;
 
-type EthersClient = Provider<RetryClient<Http>>;
+type EthersClient = Provider<RetryClient<QuorumProvider<Http>>>;
 
 /// RelayerContext contains Relayer's configuration and shutdown signal.
 #[derive(Clone)]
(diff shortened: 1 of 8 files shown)


config/README.md
```diff
@@ -327,11 +327,17 @@ The HTTP(s) RPC endpoint for this chain, used for watching events, and sending t
...
 Example:
-
+- Single Endpoint
 ```toml
 http-endpoint = "https://mainnet.infura.io/v3/<project-id>"
 ```
 
+- Multiple Endpoints
+```toml
+http-endpoint = ["https://mainnet.infura.io/v3/<project-id>","https://rpc.testnet.network"]
```
drewstone, line 337: Is it intended these multiple RPCs should be for the same network? We should include more details in the docs above if so. The type can also be an array of strings.

crates/relayer-context/src/lib.rs
```diff
@@ -120,16 +120,35 @@ impl RelayerContext {
         // Create a Map for all EVM Chains
         let mut evm_providers = HashMap::new();
         for (_, chain_config) in config.evm.iter() {
-            let client = Http::new(chain_config.http_endpoint.clone());
+            let mut providers = Vec::new();
+            match chain_config.http_endpoint.clone() {
+                webb_relayer_config::evm::HttpEndpoint::Single(rpc_url) => {
+                    let client = Http::new(rpc_url);
```
drewstone, line 126: The two blocks of code can be de-duplicated with a closure that pushes the provider to the Vec.

config/README.md
```diff
@@ -327,11 +327,17 @@ The HTTP(s) RPC endpoint for this chain, used for watching events, and sending t
...
+```
+
+
 #### ws-endpoint
```
drewstone, line 341: Should we plan to do the same with websocket endpoints too? @shekohex

drewstone (changes requested):

config/README.md
salman01zp: Yes, multiple RPCs should be for the same network. I will include more details in doc.

crates/relayer-utils/src/multi_provider.rs
```diff
@@ -0,0 +1,83 @@
...
+use serde::{de::DeserializeOwned, Serialize};
+use std::sync::{
+    atomic::{AtomicUsize, Ordering},
+    Arc,
+};
+use webb::evm::ethers::providers::{JsonRpcClient, ProviderError};
+/// MultiProvider is a JsonRpcClient that will round-robin requests to the underlying providers.
+#[derive(Debug, Clone)]
+pub struct MultiProvider<P> {
+    providers: Vec<Arc<P>>,
```
shekohex, line 13: The Arc should be wrapping the Vec, not the other way around.

shekohex: which would allow us cloning the whole thing without creating a new vec of providers.

shekohex (changes requested):
Looks good! lets update the docs too.

shekohex (approved):
Looks good! :+1:

drewstone (approved):

PR #540, opened 2023-06-20
Add polling interval configuration for providers

### LINKED ISSUE #520: [TASK] Investigate optimization/multi-call possibilities on the relayer.
# Overview
We are exhausting our INFURA api each day, quite early every day. We need to continue to optimize the RPC calls we are making. However we are fetching data, we should find new ways of consolidating these calls.

The most common calls are:
- getTxByHash
- getLogs
- etc.

## Checklist
- [x] Investigate where we can leverage multi-call
- [x] Investigate if filtering for ALL events is causing a blowup in RPC calls, if so, optimize by fetching only the events we handle
- [ ] Document ways of using multi-call for those RPC calls
- [ ] Create tasks to implement them.

### LINKED ISSUE #528: [TASK] Investigate why we call `eth_getTransactionByHash` RPC so many times
### Overview

Using Infura Stats, we noticed that we call `eth_getTransactionByHash` a lot, which is indication that there is a room left here to improve the number of RPCs we send.
image

### Task Checklist

- [ ] Explore a way to get metrics from ethers-rs (using the tracing logs for example)
- [ ] Propose a way to lower the number of RPCs
- [ ] Add tests (if required)

salman01zp:
## Summary of changes
- Add polling interval configuration for providers.
This configuration is used to define the time interval for checking pending transaction status. Higher the time interval lower RPC call will be made for `eth_getTransactionByHash` for checking transaction status

For the `eth_getLogs ` RPC call only optimization possible is to configure a high max step value for fetching events from large range of blocks

```rust
    /// The maximum number of events to fetch in one request.
    #[serde(default = "defaults::max_blocks_per_step")]
    pub max_blocks_per_step: u64,
```

- Closes #528 
- Closes #520 

-----
### Code Checklist 

- [x] Tested
- [x] Documented

### DIFF
--- a/tests/lib/localTestnet.ts
+++ b/tests/lib/localTestnet.ts
@@ -61,6 +61,7 @@ export type ExportedConfigOptions = {
   privateKey?: string;
   smartAnchorUpdates?: SmartAnchorUpdatesConfig;
   httpEndpoints?: string[];
+  txQueueConfig?: TxQueueConfig;
 };
 
 // Default Events watcher for the contracts.
@@ -69,6 +70,18 @@ export const defaultEventsWatcherValue: EventsWatcher = {
   pollingInterval: 1000,
   printProgressInterval: 60_000,
 };
+
+// Default configuration for Evm Tx queue
+export const defaultEvmTxQueueConfig: TxQueueConfig = {
+  maxSleepInterval: 10000,
+  pollingInterval: 15000,
+};
+
+// Tx queue configurations
+export interface TxQueueConfig {
+  maxSleepInterval: number;
+  pollingInterval: number;
+}
 
 type LocalChainOpts = {
   name: string;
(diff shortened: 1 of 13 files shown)


tests/lib/localTestnet.ts
```diff
@@ -70,6 +71,18 @@ export const defaultEventsWatcherValue: EventsWatcher = {
   printProgressInterval: 60_000,
 };
 
+// Default configuration for Evm Tx queue
+export const defaultEvmTxQueueConfig: TxQueueConfig = {
+  maxSleepInterval: 10000,
```
shekohex, line 76: For tests, lets make things faster, since the RPCs are running locally anyway. Block time is 1s anyway.

shekohex (approved):
Looks good.

PR #541, opened 2023-06-22
Improve queue to track item progress

### LINKED ISSUE #452: [TASK] Improve the Internal Queue System for the relayer
### Overview

While working on the docs for the Internal Queue system #390 I noticed that we could improve the queue system so that we can keep track of the item status inside the queue so that the caller could potentially depending on the status, if it is a recoverable error they can retry the item again.

### Checklist

- [x] Refactor the Queue so that when we enqueue an item it returns a unique `ItemKey`.
- [x] The caller of the Dequeue on the queue is responsible to update the Status of the item using its `ItemKey`.
- [x] if the item failed to be processed, the processor function sets its status to `Status::Failed` with the error information.
- [ ] The observer who enqueued the item should watch the updates on the `ItemKey`.
- [ ] it is up to the observer to retry the item again if the error is recoverable for example.
- [x] Add happy tests and error tests to the new queue changes.
- [x] Update the docs related to the old methods.

salman01zp:
## Summary of changes
Update Queue system to include item status tracking and TTL feature. Items with pending status will only be processed and will be removed once TTL duration expires
```rust
pub struct QueueItem<T> {
    /// The inner value wrapped by the Queue Item.
    inner: T,
    /// The current state of the item in the queue.
    state: QueueItemState,
    /// The time when the item was enqueued.
    enqueued_at: u128,
    /// Time to live
    ttl: u128,
}

/// The status of the item in the queue.
#[derive(Debug, Clone, PartialEq, Default, Serialize, Deserialize)]
pub enum QueueItemState {
    /// The current item is pending and waiting in the queue to be dequeued and processed.
    #[default]
    Pending,
    /// The item is being processed.
    Processing {
        /// A meaningful step for the current item state.
        step: String,
        /// A meaningful progress percentage for the current item state (0 to 1).
        progress: Option<f32>,
    },
    /// The item failed to be processed.
    Failed {
        /// The error message.
        reason: String,
    },
    /// The item was successfully processed.
    Processed,
}

```

- Closes #452 

-----
### Code Checklist 

- [ ] Tested
- [ ] Documented

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -33,7 +33,6 @@ jobs:
         uses: Swatinem/rust-cache@v2
         with:
           shared-key: "rust"
-          cache-directories: ".dvc/tmp\n.dvc/cache"
 
       - name: Build Checking on ${{ matrix.target }}
         run: cargo check --target=${{ matrix.target }} --features native-tls/vendored --locked
@@ -65,7 +64,6 @@ jobs:
         uses: Swatinem/rust-cache@v2
         with:
           shared-key: "rust"
-          cache-directories: ".dvc/tmp\n.dvc/cache"
 
       - name: Testing on ${{ matrix.target }}
         run: cargo test --target=${{ matrix.target }} --locked
@@ -133,7 +131,6 @@ jobs:
         uses: Swatinem/rust-cache@v2
         with:
           shared-key: "rust"
-          cache-directories: ".dvc/tmp\n.dvc/cache"
 
       - name: Create Env file
         run: |
@@ -165,6 +162,8 @@ jobs:
 
       - name: Rust Cache
         uses: Swatinem/rust-cache@v2
+        with:
+          shared-key: "rust"
         
       - name: Download fixtures
         run: |
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -163,14 +163,16 @@ jobs:
         with:
           fetch-depth: 50
 
-      - name: Setup DVC
-        uses: iterative/setup-dvc@v1
-
-      - name: Rust Cache
-        uses: Swatinem/rust-cache@v2
-        with:
-          shared-key: "rust"
-          cache-directories: ".dvc/tmp\n.dvc/cache"
+      - name: Download fixtures
+        run: |
+          mkdir -p ./tests/solidity-fixtures/vanchor_2/2
+          mkdir -p ./tests/solidity-fixtures/vanchor_16/2
+          curl -L -o ./tests/solidity-fixtures/vanchor_2/2/witness_calculator.cjs https://dapp-fixtures.s3.amazonaws.com/develop/vanchor_2/2/witness_calculator.cjs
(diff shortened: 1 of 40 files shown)


salman01zp:
Will create separate PR to process `private_tx` withdraw through `txqueue`  and with expose endpoint for item status tracking for Evm and Substrate chain

shekohex:
Will review shortly.

.github/workflows/check.yml
```diff
@@ -163,14 +163,16 @@ jobs:
...
           fetch-depth: 50
 
-      - name: Setup DVC
-        uses: iterative/setup-dvc@v1
-
-      - name: Rust Cache
-        uses: Swatinem/rust-cache@v2
-        with:
-          shared-key: "rust"
-          cache-directories: ".dvc/tmp\n.dvc/cache"
```
shekohex, line 173: Why we removed caching from here?

crates/relayer-store/src/queue.rs
```diff
@@ -0,0 +1,217 @@
...
+    pub fn new(inner: T) -> Self {
+        let now = std::time::SystemTime::now()
+            .duration_since(std::time::UNIX_EPOCH)
+            .expect("Time went backwards");
+
+        Self {
+            inner,
+            state: Default::default(),
+            enqueued_at: now.as_millis(),
+            ttl: 3 * 60 * 60 * 1000, // 3 hours
```
shekohex, line 43: Could you write a unit test with a small ttl to test out the expiry.

crates/relayer-store/src/queue.rs
```diff
@@ -0,0 +1,217 @@
...
+    /// To update an item in the queue, you MUST provide the [`QueueKey::item_key`].
+    /// Otherwise, the implementation does not know which item to update.
+    fn update_item<F>(&self, key: Self::Key, f: F) -> crate::Result<bool>
+    where
+        F: FnMut(&mut QueueItem<Item>) -> crate::Result<()>;
+
+    /// Shift item to the end of the queue.
+    fn shift_item_to_end<F>(&self, key: Self::Key, f: F) -> crate::Result<bool>
+    where
+        F: FnMut(&mut QueueItem<Item>) -> crate::Result<()>;
```
shekohex, line 137: Could you update the docs to explain the expected behavior of this method? like what is the difference between this method and `enqueue` since both add the item to the end of the queue.

shekohex (changes requested):
Looks good, need to fix the CI and address these comments.

.github/workflows/check.yml
salman01zp: Just thought we can use and delete it once done.. It takes ~5sec to download fixtures since its small

shekohex: This does not only caches the fixtures, also it does cache the rust's build artifacts from the `target` dir.

salman01zp: I see ,will add it back.

shekohex (changes requested):

PR #547, opened 2023-06-26
Implement EVM Proposals Batching

### LINKED ISSUE #538: [TASK] Implement the batching logic necessary for batching support on DKG
# Overview
We have batching implemented on the pallet side, which means that oftentimes `Batch<Proposal>` batches are signed containing more than 1 proposal. For this to completely land, we need to update the relayer to listen for batches and process them against the right transaction on the `SignatureBridge.sol`.

### Contract
https://github.com/webb-tools/protocol-solidity/blob/main/packages/contracts/contracts/SignatureBridge.sol#L144-L153

### Nuances
The functions in question are:
- `batchAdminSetResourceWithSignature`
- `batchExecuteProposalsWithSignature`

We need to ensure that `SetResourceProposal` is only batched together with other `SetResourceProposal`s. It cannot be batched with an `AnchorUpdateProposal` for example. This is due to the fact that we handle it exclusively on the `SignatureBridge.sol` in an entirely separate fashion. For example if we `batchExecuteProposalsWithSignature` or `executeProposalWithSignature` with a `SetResourceProposal` nothing will happen because there is no "handler" for this proposal type.

## Event handling
The batch signed event is new and should be handled as such by the relayer, `ProposalBatchSigned`.

## Checklist
- [x] Update contracts on webb-rs to new 1.0.0+ versions
- [x] Update the relayer to leverage this new function `batchAdminSetResourceWithSignature` and `batchExecuteProposalsWithSignature` (do match statement over the batch size).
- [ ] If the batch contains 1 proposal, consider just calling the normal original function (`executeProposalWithSignature`, `adminSetResourceWithSignature`)
- [ ] We need to handle `SetResourceProposals` properly and currently we aren't at all. Parse the proposal type in order to switch between the set resource flow and the execute proposal flow.

shekohex:
## Summary of changes
- add handler for `batchAdminSetResourceWithSignature`.
- add handler for `batchExecuteProposalsWithSignature`.
- Handle `ProposalBatchSigned` event.

- Closes #538 

-----
### Code Checklist 

- [ ] Tested
- [x] Documented


shekohex:
Tests will fail ofc, since the setup for batching is not included in the main tangle, yet. we are going to test locally for now.

crates/relayer-utils/src/static_tx_payload.rs
```diff
@@ -57,16 +60,31 @@ impl<CallData: Encode> TryFrom<StaticTxPayload<CallData>>
...
         pallet_index.encode_to(out);
         call_index.encode_to(out);
-        out.extend_from_slice(&self.call_data);
+
+        let mut fields = call
+            .fields
+            .iter()
+            .map(|f| scale_encode::Field::new(f.ty.id, f.name.as_deref()));
+
+        self.call_data.encode_to(out);
```
shekohex, line 87: will need to figure out if we actually ever need this.

PR #565, opened 2023-07-24
Use Insertion Event instead of NewCommitment

### LINKED ISSUE #564: [TASK] Genrialze EVM Event Watchers over `MerkleSystem`s instead of VAnchors
## Task Overview

Looks like #500 did not actually fix the root issue with the Syncing leaves and verifying that the new leaves/commitments are valid always works, I discovered that this bug is still there as I was recently testing Webb Alpha dApp with the current Relayers that they use [v0.5.0-rc23-alpha](https://github.com/webb-tools/relayer/releases/tag/v0.5.0-rc23-alpha) which has that bug fix for a while now. But as you may know that some of these nodes just returns the default root since it does not keep the full history, even tho we query the root at specific block number at the time that event of `NewCommitment` occurred. However some RPCs does only keep the last 250 blocks of state, maybe Infura keeps more than that, but anyway looks like querying the merkle root separately from the event does not work very well. 

During the discussion while working on #500, we mentioned emitting the current merkle root along with the commitment itself in one event, this way we can verify and validate the merkle tree without another query to the RPC. During the PR https://github.com/webb-tools/protocol-solidity/pull/328/ We noticed a change that [`Insertion`](https://github.com/webb-tools/protocol-solidity/blob/848d073bb17f0aaffc6d39f594cc59efedeaec89/packages/contracts/contracts/interfaces/IMerkleSystem.sol#L60-L65) Now Includes `newMerkleRoot` field which we can utilize to build the merkle tree instead.

## Task Checklist

- [ ] Change the handlers to use `Insertion` Event instead of `NewCommitment` Event
- [ ] Update tests
- [ ] Pump the relayer version
- [ ] Create a new release.

shekohex:
## Summary of changes
- Start using `Insertion` event for handling new commitment for the leaves watchers.
- Update the logic for saving the merkle-tree validation.

- Closes #564 

-----
### Code Checklist 

- [x] Tested
- [x] Documented

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9500,7 +9500,7 @@ dependencies = [
 
 [[package]]
 name = "webb-bridge-registry-backends"
-version = "0.5.7-dev"
+version = "0.5.8-dev"
 dependencies = [
  "async-trait",
  "ethereum-types 0.14.1",
@@ -9514,12 +9514,12 @@ dependencies = [
  "webb 0.7.3",
  "webb-proposals",
  "webb-relayer-config",
- "webb-relayer-utils 0.5.7-dev",
+ "webb-relayer-utils 0.5.8-dev",
 ]
 
 [[package]]
 name = "webb-chains-info"
-version = "0.5.7-dev"
+version = "0.5.8-dev"
 dependencies = [
  "anyhow",
  "prettyplease",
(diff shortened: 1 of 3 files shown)


drewstone (approved):
LGTM

PR #571, opened 2023-08-09
Add relayer fee configuration to relayer

### LINKED ISSUE #570: [TASK] Make Private Transaction Relayer Fees configurable
Currently hard-coded to $5:
https://github.com/webb-tools/relayer/blob/1b020b8cd639f6e4de441995d387f762f12f6401/crates/tx-relay/src/lib.rs#L8

We need it to be $% instead of flat $N for example, if the tx fees is $1 and our profit is set to 10% then we will charge $1.01 fees. the value could be over 100%, so it is just `f64` in percent. This value should be configurable for each network.

Same goes for Max Refund value, hardcoded here:
https://github.com/webb-tools/relayer/blob/1b020b8cd639f6e4de441995d387f762f12f6401/crates/tx-relay/src/lib.rs#L5-L6

We need to make it more configurable for each network.

salman01zp:
## Summary of changes
- [x] Add Relayer fee configuration
- [x] Update test
- [x] Update config files
- [x] Update readme

Added Relayer fee configuration, which can be used by each chain to configure `relayer_profit_percent` and  `max_refund_amount` of the relayer

```rust
/// Transaction relayer fee configuration.
#[derive(Debug, Clone, Deserialize, Serialize)]
#[serde(rename_all(serialize = "camelCase", deserialize = "kebab-case"))]
pub struct RelayerFeeConfig {
    /// Relayer profit percent per transaction fee for relaying
    pub relayer_profit_percent: f64,
    /// Maximum refund amount per transaction relaying
    pub max_refund_amount: f64,
}

impl Default for RelayerFeeConfig {
    fn default() -> Self {
        Self {
            relayer_profit_percent: 5.,
            max_refund_amount: 5.,
        }
    }
}
```

- Closes #570 

-----
### Code Checklist 

- [ ] Tested
- [ ] Documented

### DIFF
--- a/config/README.md
+++ b/config/README.md
@@ -33,6 +33,9 @@ following section we will describe the different configuration entries and how t
   - [beneficiary](#beneficiary)
   - [tx-queue](#tx-queue)
     - [max-sleep-interval](#max-sleep-interval)
+  - [relayer-fee-config](#relayer-fee-config)
+    - [relayer-profit-percent](#relayer-profit-percent)
+    - [max-refund-amount](#max-refund-amount)
   - [contracts](#contracts)
     - [contract](#contract)
     - [address](#address)
@@ -467,6 +470,29 @@ Example:
 
 ```toml
 tx-queue = { max-sleep-interval = 5000, polling-interval = 12000 }
+```
+
+#### Relayer fee Config
+This configuration is used to configure profit percentage margin for relayer and set maximum amount relayer can refund to user while withdrawing.
+
+##### relayer-profit-percent
+The percentage of profit relayer will get from the transaction fee.
+
+- Type: `number`
+- Required: `false`
+- Default: `5`
+
+##### max-refund-amount
+The maximum amount of native token that relayer will refund to the user.
+
+- Type: `number`
+- Required: `false`
+- Default: `5`
+
+Example:
+
+```toml
+relayer-fee-config = { relayer-profit-precent = 5, max-refund-amount = 5 }
 ```
 
 #### Contracts
--- a/config/README.md
+++ b/config/README.md
@@ -492,7 +492,7 @@ The maximum amount of native token that relayer will refund to the user.
 Example:
 
 ```toml
-relayer-fee-config = { relayer-profit-precent = 5, max-refund-amount = 5 }
+relayer-fee-config = { relayer-profit-percent= 5, max-refund-amount = 5 }
 ```
 
 #### Contracts
--- a/config/README.md
+++ b/config/README.md
@@ -483,7 +483,7 @@ The percentage of profit relayer will get from the transaction fee.
 - Default: `5`
 
 ##### max-refund-amount
-The maximum amount of native token that relayer will refund to the user.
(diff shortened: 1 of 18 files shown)


crates/relayer-config/src/evm/mod.rs
```diff
@@ -71,11 +71,33 @@ pub struct EvmChainConfig {
     /// TxQueue configuration
     #[serde(skip_serializing, default)]
     pub tx_queue: TxQueueConfig,
+    /// Relayer fee configuration
+    #[serde(skip_serializing, default)]
```
shekohex, line 75: I would expose this to the api, so the dapp can sort the relayers by fee percentage

crates/tx-relay/src/evm/fees.rs
```diff
@@ -236,7 +260,9 @@ fn calculate_transaction_fee(
     let tx_fee_usd = tx_fee_tokens * native_token_price;
     // Step 3: Calculate the profit that the relayer should make, and add it to the tx fee in USD.
     // This is the total amount of USD that the relayer should receive.
-    let total_fee_with_profit_in_usd = tx_fee_usd + TRANSACTION_PROFIT_USD;
+    let relay_tx_fee =
```
shekohex, line 263:
```suggestion
    let relay_tx_profit =
```

crates/tx-relay/src/evm/fees.rs
```diff
@@ -236,7 +260,9 @@ fn calculate_transaction_fee(
...
+        (relayer_fee_config.relayer_profit_percent / 100.0) * tx_fee_usd;
+    let total_fee_with_profit_in_usd = tx_fee_usd + relay_tx_fee;
```
shekohex, line 265:
```suggestion
    let total_fee_with_profit_in_usd = tx_fee_usd + relay_tx_profit;
```

config/README.md
```diff
@@ -469,6 +472,29 @@ Example:
...
+
+##### relayer-profit-percent
+The percentage of profit relayer will get from the transaction fee.
+
+- Type: `number`
+- Required: `false`
+- Default: `5`
+
+##### max-refund-amount
+The maximum amount of native token that relayer will refund to the user.
```
shekohex, line 486: This inaccurate, as it's the max amount of native tokens in USD. I would actually change the logic rather than changing the docs.

shekohex (changes requested):
Looks good so far.

config/README.md
salman01zp: yes. I think we should keep the logic as it was to return x USD amount of native tokens instead of x native tokens.
 Since users need just a small amount and is easy to calculate in USD

shekohex: Cool.

shekohex (approved):
Looks good, last thing is to update the relayer version to v0.5.10-dev so we can release a new version

shekohex:
@salman01zp any idea why the CI is failing?

salman01zp:
hmm, some issue with the tangle node.
https://github.com/webb-tools/relayer/actions/runs/5798481431/job/15719574758#step:10:66

PR #580, opened 2023-09-14
Cache substrate rpc

### LINKED ISSUE #579: [TASK] Optmize Substrate Connections
### Overview
Recently, we found a bug that spams (opens a lot of) connections to the tangle archival nodes, this may be due to a lot of clients that gets created instead of cached. we found it easily identifiable from the logging here:
 https://github.com/webb-tools/relayer/blob/f42ec552697faca1f6bd3b5749ce26c11d702d66/crates/event-watcher-traits/src/substrate/event_watcher.rs#L159-L171 which was introduced through this PR #522 

Every call to [`ctx.substrate_provider`](https://github.com/webb-tools/relayer/blob/f42ec552697faca1f6bd3b5749ce26c11d702d66/crates/relayer-context/src/lib.rs#L222) actually creates a new connection (even for the same chain id) which is a waste of resources, what we should do instead is to cache the RPC clients in a hashmap so that we can reuse them across different clients. That is exactly we did [way back](https://github.com/webb-tools/relayer/commit/17dc90c56c1ae8ba122b91abe505f2370eab37bc) for EVM Providers.

### Task Checklist

- [x] Create a shared RPC Client and store it in `Arc<HashMap<types::U256, Arc<RpcClient>>>`
- [x] Construct the RPC Clients during the `RelayerContext::new` call (notice that creating the clients does not actually connect)
- [x] Whenever we call `RelayerContext::substrate_provider` method, we create a new `OnlineClient` from the shared `Arc<RpcClient>` we cached.
- [x] if this failed, calling this again with a same client will reuse the connection.
- [x] Can we use `HttpClient` as RPC Client? if yes, maybe that is a better option as they are stateless.
- [ ] (_optional_) While constructing the RpcClient, please add `User-Agent` header to `webb-relayer/<version> <build-commit>` so that we can see these in the Graphana logs/UI.

### References

1. Creating a custom RPC Client: https://github.com/paritytech/subxt/blob/c8462defabad10a2c09f945737731e7259f809dd/subxt/src/backend/rpc/rpc_client.rs#L240-L253
2. Using the created RPC Client to create an OnlineClient: https://github.com/paritytech/subxt/blob/059723e4313d91e8ca0bcd84b0dd7dd66686ca50/subxt/src/client/online_client.rs#L86

salman01zp:
## Summary of changes
- 

- Closes #579 

-----
### Code Checklist 

- [ ] Tested
- [ ] Documented

### DIFF
--- a/crates/relayer-types/Cargo.toml
+++ b/crates/relayer-types/Cargo.toml
@@ -11,6 +11,7 @@ repository = { workspace = true }
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+webb-relayer-utils = { workspace = true }
 tracing = { workspace = true }
 url = { workspace = true }
 serde = { workspace = true }
@@ -18,6 +19,9 @@ webb = { workspace = true }
 # Used by ethers (but we need it to be vendored with the lib).
 native-tls = { workspace = true }
 ethereum-types = { workspace = true }
-
+jsonrpsee = { workspace = true, features = ["macros",
+  "ws-client",
+  "http-client"]}
+http = "0.2.9"
 subxt-signer = { workspace = true }
 tiny-bip39 = "1.0.0"
--- a/crates/relayer-types/Cargo.toml
+++ b/crates/relayer-types/Cargo.toml
@@ -19,9 +19,7 @@ webb = { workspace = true }
 # Used by ethers (but we need it to be vendored with the lib).
 native-tls = { workspace = true }
 ethereum-types = { workspace = true }
-jsonrpsee = { workspace = true, features = ["macros",
-  "ws-client",
-  "http-client"]}
+jsonrpsee = { workspace = true, features = ["macros","ws-client"]}
 http = "0.2.9"
 subxt-signer = { workspace = true }
 tiny-bip39 = "1.0.0"
--- a/crates/relayer-types/Cargo.toml
+++ b/crates/relayer-types/Cargo.toml
@@ -19,7 +19,7 @@ webb = { workspace = true }
 # Used by ethers (but we need it to be vendored with the lib).
 native-tls = { workspace = true }
 ethereum-types = { workspace = true }
-jsonrpsee = { workspace = true, features = ["macros","ws-client"]}
+jsonrpsee = { workspace = true, features = ["ws-client"]}
 http = "0.2.9"
 subxt-signer = { workspace = true }
 tiny-bip39 = "1.0.0"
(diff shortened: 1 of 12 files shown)


crates/relayer-types/Cargo.toml
```diff
@@ -11,13 +11,17 @@ repository = { workspace = true }
...
 url = { workspace = true }
 serde = { workspace = true }
 webb = { workspace = true }
 # Used by ethers (but we need it to be vendored with the lib).
 native-tls = { workspace = true }
 ethereum-types = { workspace = true }
-
+jsonrpsee = { workspace = true, features = ["macros",
+  "ws-client",
+  "http-client"]}
```
shekohex, line 24: Do we need macros, http-client features? if not we should remove them. plus the formatting.

crates/relayer-types/src/rpc_client.rs
```diff
@@ -0,0 +1,50 @@
+use std::sync::Arc;
+
+pub use jsonrpsee::core::client::Client;
+use webb::substrate::subxt::{self, rpc::RpcClientT};
+
+#[derive(Debug, Clone)]
+pub struct WebbRpcClient(pub Arc<jsonrpsee::async_client::Client>);
```
shekohex, line 7: Do we need an inner `Arc` here? we have an `Arc` around the `WebbRpcClient` already, lets remove the inner Arc around `jsonrpsee::async_client::Client`.

crates/relayer-types/src/rpc_client.rs
```diff
@@ -0,0 +1,50 @@
...
+            webb_relayer_utils::Error::Generic("RPC url is invalid")
+        })?;
+        let (sender, receiver) =
+            jsonrpsee::client_transport::ws::WsTransportClientBuilder::default(
+            )
+            .build(url)
+            .await
+            .map_err(|_| {
+                webb_relayer_utils::Error::Generic("RPC failed to connect")
+            })?;
```
shekohex, line 23: maybe you should import the `jsonrpsee::client_transport::ws::WsTransportClientBuilder` instead of the fully qualified path here? code looks a bit ugly and rustfmt does not help it.

shekohex (changes requested):

crates/relayer-types/src/rpc_client.rs
salman01zp: I made it make `WebbRpcClient` clonable, but since we are wrapping it in Arc so we can remove it..

shekohex (approved):

